PR #1, opened 2024-01-18
Enable CI

peter-marcisovsky:
Adding `.gitlab` folder to enable CI on newly migrated esp-usb repository

### DIFF
--- a/.github/workflows/build_and_run_test_app_usb.yml
+++ b/.github/workflows/build_and_run_test_app_usb.yml
@@ -1,0 +1,65 @@
+name: Build and Run USB Test Application
+
+on:
+  schedule:
+    - cron: '0 0 * * SAT' # Saturday midnight
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+jobs:
+  build:
+    name: Build USB Test App
+    strategy:
+      fail-fast: false
+      matrix:
+        idf_ver: ["release-v4.4", "release-v5.0", "release-v5.1", "release-v5.2", "latest"]
+    runs-on: ubuntu-20.04
+    container: espressif/idf:${{ matrix.idf_ver }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: 'true'
+      - name: Build USB Test Application
+        shell: bash
+        run: |
+          . ${IDF_PATH}/export.sh
+          pip install idf-component-manager idf-build-apps --upgrade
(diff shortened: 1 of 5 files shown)


.github/workflows/build_and_run_test_app_usb.yml
```diff
@@ -0,0 +1,72 @@
+name: Build and Run USB Test Application
+
+on:
+  schedule:
+    - cron: '0 0 * * SAT' # Saturday midnight
+  pull_request:
+    types: [opened, reopened, synchronize]
+    paths:
+      - 'usb/**'
+      - 'workflows/build_and_run_test_app_usb.yml'
```
tore-espressif, line 10: This section is not required anymore. In idf-extra-components we only wanted to run this pipeline only when changes it `usb` directory were made.

(The path on line 10 is incorrect, so the GH action is not run at all)

.github/workflows/build_and_run_test_app_usb.yml
```diff
@@ -0,0 +1,72 @@
...
+        env:
+          IDF_TARGET: ${{ matrix.idf_target }}
+        shell: bash
+        working-directory: usb/test_app
+        run: |
+          . ${IDF_PATH}/export.sh
+          export PEDANTIC_FLAGS="-DIDF_CI_BUILD -Werror -Werror=deprecated-declarations -Werror=unused-variable -Werror=unused-but-set-variable -Werror=unused-function"
+          export EXTRA_CFLAGS="${PEDANTIC_FLAGS} -Wstrict-prototypes"
+          export EXTRA_CXXFLAGS="${PEDANTIC_FLAGS}"
+          idf.py build
```
tore-espressif, line 36: After we are sure that the CI infrastructure works, could you please refactor this using https://github.com/espressif/idf-build-apps ?

Example here: https://github.com/espressif/esp-bsp/blob/master/.github/workflows/build_example.yml#L43 (there is no need for parallel runs for now)

.github/workflows/upload_component.yml
```diff
@@ -0,0 +1,49 @@
...
+            usb/usb_host_cp210x_vcp;
+            usb/usb_host_ftdi_vcp;
+            usb/usb_host_hid;
+            usb/usb_host_msc;
+            usb/usb_host_uvc;
+            usb/usb_host_vcp;
+          namespace: "espressif"
+          # API token will only be available in the master branch in the main repository.
+          # However, dry-run doesn't require a valid token.
+          api_token: ${{ secrets.IDF_COMPONENT_API_TOKEN }}
```
tore-espressif, line 48: I will add the `IDF_COMPONENT_API_TOKEN` and we can merge

tore-espressif (approved):

PR #4, opened 2024-01-25
usb_host_msc: Add P4 support

tore-espressif:
Fix MSC driver for P4

1. Correctly handle UTF-16 string descriptors
2. Revert zero-copy transfers. They cause issues with P4's L1 cache


roma-jam (approved):
Current speed on my device: 
I (181853) example: Writing to file /usb/esp/dummy
I (183503) example: Write speed 0.64 MiB/s
I (183503) example: Reading from file /usb/esp/dummy
I (183733) example: Read speed 4.56 MiB/s

Otherwise, LGTM

PR #5, opened 2024-01-25
usb_host_hid v1.0.2: Support ESP32-P4

roma-jam:
# usb_host_hid v1.0.2

RELEASE

## Change description
- String descriptor copying procedure doesn't use `wcsncpy` anymore. Now, there is no difference on sizeof(wchar_t) on different platforms (riscv, xtensa).
- Removed the flag `-Wno-address-of-packed-member` which was necessary for `wcsncpy`.


tore-espressif (approved):
LGTM!

PR #6, opened 2024-01-26
esp_tinyusb: Removed MIN() definition if defined for cdc_acm

roma-jam:
# esp_tinyusb v1.4.4

NOT A RELEASE 

## Change description
- Removed MIN() macro definition if already defined (CDC-ACM)
- Changed the EP size for MSC class in default configuration descriptor (for HS)

### DIFF
--- a/device/esp_tinyusb/usb_descriptors.c
+++ b/device/esp_tinyusb/usb_descriptors.c
@@ -182,7 +182,7 @@ uint8_t const descriptor_cfg_kconfig[] = {
 
 #if CFG_TUD_MSC
     // Interface number, string index, EP Out & EP In address, EP size
-    TUD_MSC_DESCRIPTOR(ITF_NUM_MSC, STRID_MSC_INTERFACE, EPNUM_MSC, 0x80 | EPNUM_MSC, 64), // highspeed 512
+    TUD_MSC_DESCRIPTOR(ITF_NUM_MSC, STRID_MSC_INTERFACE, EPNUM_MSC, 0x80 | EPNUM_MSC, TUD_OPT_HIGH_SPEED ? 512 : 64),
 #endif
 
 #if CFG_TUD_NCM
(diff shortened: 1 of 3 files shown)


tore-espressif (approved):
@roma-jam are there any outstanding issues or can we merge?

roma-jam:
@tore-espressif not for now. There will be changes in esp_tinyusb test, but if there is something urgent that requires changes in that pr - can be merged.

device/esp_tinyusb/usb_descriptors.c
```diff
@@ -182,7 +182,7 @@ uint8_t const descriptor_cfg_kconfig[] = {
 
 #if CFG_TUD_MSC
     // Interface number, string index, EP Out & EP In address, EP size
-    TUD_MSC_DESCRIPTOR(ITF_NUM_MSC, STRID_MSC_INTERFACE, EPNUM_MSC, 0x80 | EPNUM_MSC, 64), // highspeed 512
+    TUD_MSC_DESCRIPTOR(ITF_NUM_MSC, STRID_MSC_INTERFACE, EPNUM_MSC, 0x80 | EPNUM_MSC, TUD_OPT_HIGH_SPEED ? 512 : 64),
```
tore-espressif, line 185: Could we do the same for default NCM descriptor? This way we won't hit the same problem once we port HighSpeed NCM in the future

roma-jam: @tore-espressif For NCM class default configuration descriptor has already the correct size of BULK EP: 
`#define CFG_TUD_NET_ENDPOINT_SIZE (TUD_OPT_HIGH_SPEED ? 512 : 64)`
For INTR EP I am not sure that this is important, because the INTR could be any size in HS/FS, not strictly 512/64....
But let me check that additionally.

tore-espressif: Thank you! I indeed look at the INTR EP size, which is 64 and correct.

PR #8, opened 2024-02-01
esp-usb:  Add codespell pre-commit hook

roma-jam:
## Change description
- Added a codespell pre-commit hook (for `c` and `cpp` files only, with `codespell-ignore-list`)
- Applied codespell on all files (`c` and `cpp`)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,13 @@ repos:
     -   id: astyle_py
         args: ['--style=otbs', '--attach-namespaces', '--attach-classes', '--indent=spaces=4', '--convert-tabs', '--align-pointer=name', '--align-reference=name', '--keep-one-line-statements', '--pad-header', '--pad-oper']
 
+-   repo: https://github.com/codespell-project/codespell
+    rev: v2.2.6
+    hooks:
+    - id: codespell
+      types_or: [c, c++]
+      args: [-w, "--ignore-words=codespell-ignore-list"]
+
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.3.0
     hooks:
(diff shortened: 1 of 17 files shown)


.pre-commit-config.yaml
```diff
@@ -5,6 +5,13 @@ repos:
     -   id: astyle_py
         args: ['--style=otbs', '--attach-namespaces', '--attach-classes', '--indent=spaces=4', '--convert-tabs', '--align-pointer=name', '--align-reference=name', '--keep-one-line-statements', '--pad-header', '--pad-oper']
 
+-   repo: https://github.com/codespell-project/codespell
+    rev: v2.2.6
+    hooks:
+    - id: codespell
+      types_or: [c, c++]
```
tore-espressif, line 12: Would it make sense to add markdown files too? such as readmes?

tore-espressif (commented):
Thank you very much @roma-jam !

.pre-commit-config.yaml
roma-jam: Yes, sure. 
I didn't expect there will be so much typos in usb host class drivers....

I have added just the source files, but I think we can add almost everything from the project.

Dazza0 (approved):
@roma-jam Thanks for the addition. LGTM

.pre-commit-config.yaml
roma-jam: @tore-espressif Done. Now the codespell applies on all files without exceptions. I tried to find any files that we do not need to check, but I didn't come up with anything. 

Do you think there is something we **_don't need to verify_**?

tore-espressif: We don't need to verify files that are not intended for users, such as tests or repo infrastrucutre files. But in this case it might be easier to just spell check all files. LGTM!

tore-espressif (approved):

tore-espressif:
@roma-jam We forgot to add the pre-commit to CI, follow-up here #12

PR #10, opened 2024-02-05
esp_tinyusb: Adding bvalid_signal test

roma-jam:
# esp_tinyusb v1.4.4

NOT A RELEASE 

## Change description
- Added a usb_device test for `bvalid_signal` verification. 

## Details
USB Device events, such as attach (plug) and detach (unplug) are detected via `bvalid_signal`, which can be connected to any (*mostly) external PIN. (In the most common cases, this external PIN connected to VBUS USB via resistor divider to detect the VBUS power presence. For reference, check [Self-Powered Device](https://docs.espressif.com/projects/esp-idf/en/v5.1.1/esp32s2/api-reference/peripherals/usb_device.html))

When USB Device (self-powered) is attached to the USB Host, the VBUS becomes High. 
When USB Device (self-powered) is detached from the USB Host, the VBUS becomes Low. 
Both of attach and detach, trigger the dcd_dwc2 layer interrupt with `DCD_EVENT_UNPLUGGED` event. 

Current test implement the USB Device without any interface and connect the `bvalid_signal` to the High or Low signal source to emulate the VBUS presence/absence.
For smoke testing, 10 iterates is enough to see that logic is working. 
For load testing, please, change the parameter `DEVICE_DETACH_TEST_ROUNDS` in `test_bvalid_sig.c`.

### Important notice
Without changes in https://github.com/espressif/tinyusb/pull/19 the test doesn't work.

### DIFF
--- a/device/esp_tinyusb/test/test_bvalid_sig.c
+++ b/device/esp_tinyusb/test/test_bvalid_sig.c
@@ -1,0 +1,96 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: CC0-1.0
+ */
+
+#include "soc/soc_caps.h"
+
+#if SOC_USB_OTG_SUPPORTED
+
+#include <stdio.h>
+#include <string.h>
+#include "esp_system.h"
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+#include "esp_log.h"
+#include "esp_err.h"
+#include "driver/gpio.h"
+#include "esp_rom_gpio.h"
+#include "soc/gpio_sig_map.h"
+#include "unity.h"
+#include "tinyusb.h"
+#include "class/msc/msc_device.h"
+
+#define DEVICE_DETACH_TEST_ROUNDS       10
+#define DEVICE_DETACH_ROUND_DELAY_MS    1000
+
+/* TinyUSB descriptors
+   ********************************************************************* */
+#define TUSB_DESC_TOTAL_LEN         (TUD_CONFIG_DESC_LEN)
+
+static unsigned int dev_mounted = 0;
+static unsigned int dev_umounted = 0;
+
+static uint8_t const test_configuration_descriptor[] = {
+    // Config number, interface count, string index, total length, attribute, power in mA
+    TUD_CONFIG_DESCRIPTOR(1, 0, 0, TUSB_DESC_TOTAL_LEN, TUSB_DESC_CONFIG_ATT_SELF_POWERED | TUSB_DESC_CONFIG_ATT_REMOTE_WAKEUP, 100),
+};
+
+static const tusb_desc_device_t test_device_descriptor = {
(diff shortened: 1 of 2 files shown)


device/esp_tinyusb/CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
-## 1.4.4 (Unreleased)
+## 1.4.4
 
+- esp_tinyusb: Add BValid signal test
```
tore-espressif, line 3: * we usually only mention changes relevant for users. So we no release notes for tests

device/esp_tinyusb/CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
-## 1.4.4 (Unreleased)
+## 1.4.4
```
tore-espressif, line 1: Let's release after #11

device/esp_tinyusb/test/test_bvalid_sig.c
```diff
@@ -0,0 +1,96 @@
...
+    TEST_ASSERT_EQUAL(ESP_OK, tinyusb_driver_install(&tusb_cfg));
+
+    while (rounds--) {
+        // LOW to emulate disconnect USB device
+        esp_rom_gpio_connect_in_signal(GPIO_MATRIX_CONST_ZERO_INPUT, USB_SRP_BVALID_IN_IDX, false);
+        vTaskDelay(pdMS_TO_TICKS(DEVICE_DETACH_ROUND_DELAY_MS));
+        // HIGH to emulate connect USB device
+        esp_rom_gpio_connect_in_signal(GPIO_MATRIX_CONST_ONE_INPUT, USB_SRP_BVALID_IN_IDX, false);
+        vTaskDelay(pdMS_TO_TICKS(DEVICE_DETACH_ROUND_DELAY_MS));
+    }
```
tore-espressif, line 88: There were users complaining about disconnection events errors while there were ongoing transfers.

Could you please make this device a 'real' device with e.g. CDC class that keeps sending data to the host? And then mimic disconnection while the transfers are in progress

tore-espressif (changes requested):
@roma-jam Thank for the test! Could you please add another test that would the same as `bvalid_signal` but with ongoing transfers?

Also please consider renaming to `sudden disconnection`... :)

device/esp_tinyusb/test/test_bvalid_sig.c
roma-jam: @tore-espressif, sure, no problem. 
This will be another task and another idea though. Current changes fixes the bvalid_signal override to get the IRQ event. 

During the test, we keep the device attached, but emulate the disconnection. 
In real life, detach is detach, and when the device is gone - sure, the could be a interrupted transfer. 
IMHO, if the idea - is to recover correctly after sudden disconnection, it is better to create another ticket for that. 

Do you have any link to the complaining about error during disconnection?

tore-espressif: Only reports from AE team.

OK, please let us know if you plan any more work on this PR so we can go ahead with the merging.

PR #11, opened 2024-02-06
esp_tinyusb v1.4.4: Adding FS configuration descriptor for HS device

roma-jam:
# esp_tinyusb v1.4.4 (Release)

## General description
- USB Device now able to work with both Hosts: High Speed (HS) or Full Speed (FS)
 
## Changes
- Moved all the descriptor preparation logic inside `descriptor_control.c` file, `tinyusb_set_descriptors(..)`. This function is responsible for descriptors preparation based on the tinyusb configuration, provided during tinyusb driver install.
- Renamed default descriptors from `*_kconfig` to `*_default`.
- Added `descriptor_hs_cfg_default[]` for HighSpeed device and  `descriptor_hs_cfg_default[]` for FullSpeed device.
- Added `.hs_cfg_desc` field to tinyusb configuration (only when `TUD_OPT_HIGH_SPEED` is set)
- Added `.qualifier_desc` field to tinyusb configuration (only when `TUD_OPT_HIGH_SPEED` is set)

## Details
Configuration descriptor definitions: 
1. **FS only**
    1.1 Configuration descriptor NOT present (`.configuration_descriptor = NULL`). Default FS compatible configuration descriptor is using.
    1.2 Configuration descriptor PRESENT (`.configuration_descriptor != NULL`). Provided descriptor is using. 
2. **HS/FS**  
    2.1 Configuration descriptor NOT present (`.configuration_descriptor = NULL`). Default FS compatible configuration descriptor is using.
    2.2 HS configuration descriptor NOT present (`.hs_cfg_desc = NULL`). Default HS compatible configuration descriptor is using. 
    2.3 Configuration descriptor PRESENT (`.configuration_descriptor != NULL`). Provided FS descriptor is using. 
    2.4 HS configuration descriptor PRESENT (`.hs_cfg_desc != NULL`). Provided HS descriptor is using.

IMPORTANT: In any cases for HS/FS, configuration descriptors **must be equal** size.   

**Hints** 
- For HS/FS field `.fs_cfg_desc` might be used instead of `.configuration_descriptor`
- In HS all combinations are possible (2.1/2.2, 2.1/2.4, 2.3/2.2, 2.3/2.4), but HS and FS configuration descriptors must be the same size.
- Device qualifier descriptor must be present if device descriptor is not default. Otherwise, default qualifier descriptor is using.

### DIFF
--- a/device/esp_tinyusb/CHANGELOG.md
+++ b/device/esp_tinyusb/CHANGELOG.md
@@ -1,27 +1,28 @@
-## 1.4.4 (Unreleased)
+## 1.4.4
 
-- CDC-ACM: Remove MIN() definition if already defined
-- MSC: Set EP size in configuration descriptor based on speed
+- esp_tinyusb: Added HighSpeed and Qualifier device descriptors in tinyusb configuration
+- CDC-ACM: Removed MIN() definition if already defined
+- MSC: Fixed EP size selecting in default configuration descriptor
 
 ## 1.4.3
 
-- esp_tinyusb: ESP32P4 HS only support
+- esp_tinyusb: Added ESP32P4 support (HS only)
 
 ## 1.4.2
 
-- MSC: Fix maximum files open
-- Add uninstall function
+- MSC: Fixed maximum files open
+- Added uninstall function
 
 ## 1.4.0
 
-- MSC: Fix integer overflows
-- CDC-ACM: Remove intermediate RX ringbuffer
-- CDC-ACM: Increase default FIFO size to 512 bytes
-- CDC-ACM: Fix Virtual File System binding
+- MSC: Fixed integer overflows
+- CDC-ACM: Removed intermediate RX ringbuffer
+- CDC-ACM: Increased default FIFO size to 512 bytes
+- CDC-ACM: Fixed Virtual File System binding
 
 ## 1.3.0
 
-- Add NCM extension
+- Added NCM extension
 
 ## 1.2.1 - 1.2.2
 
@@ -29,15 +30,15 @@
 
 ## 1.2.0
 
-- Add MSC extension for accessing SPI Flash on memory card https://github.com/espressif/idf-extra-components/commit/a8c00d7707ba4ceeb0970c023d702c7768dba3dc
+- Added MSC extension for accessing SPI Flash on memory card https://github.com/espressif/idf-extra-components/commit/a8c00d7707ba4ceeb0970c023d702c7768dba3dc
 
 ## 1.1.0
 
-- Add support for NCM, ECM/RNDIS, DFU and Bluetooth TinyUSB drivers https://github.com/espressif/idf-extra-components/commit/79f35c9b047b583080f93a63310e2ee7d82ef17b
+- Added support for NCM, ECM/RNDIS, DFU and Bluetooth TinyUSB drivers https://github.com/espressif/idf-extra-components/commit/79f35c9b047b583080f93a63310e2ee7d82ef17b
 
 ## 1.0.4
 
-- Clean up string descriptors handling https://github.com/espressif/idf-extra-components/commit/046cc4b02f524d5c7e3e56480a473cfe844dc3d6
+- Cleaned up string descriptors handling https://github.com/espressif/idf-extra-components/commit/046cc4b02f524d5c7e3e56480a473cfe844dc3d6
 
 ## 1.0.2 - 1.0.3
 
--- a/device/esp_tinyusb/descriptors_control.c
+++ b/device/esp_tinyusb/descriptors_control.c
@@ -1,20 +1,44 @@
 /*
- * SPDX-FileCopyrightText: 2020-2023 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2020-2024 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
 
 #include <string.h>
 #include "esp_log.h"
+#include "esp_check.h"
+#include "esp_err.h"
 #include "descriptors_control.h"
+#include "usb_descriptors.h"
 
-#define USB_STRING_DESCRIPTOR_ARRAY_SIZE 8 // Max 8 string descriptors for a device. LANGID, Manufacturer, Product, Serial number + 4 user defined
 #define MAX_DESC_BUF_SIZE 32               // Max length of string descriptor (can be extended, USB supports lengths up to 255 bytes)
 
 static const char *TAG = "tusb_desc";
-static const tusb_desc_device_t *s_device_descriptor;
-static const uint8_t *s_configuration_descriptor;
-static const char *s_str_descriptor[USB_STRING_DESCRIPTOR_ARRAY_SIZE]; // Array of pointers (strings). Statically allocated, can be made dynamic in the future.
+
+// =============================================================================
+// STRUCTS
+// =============================================================================
+
+/**
+ * @brief Descriptor pointers for tinyusb descriptor requests callbacks
+ *
+ */
+typedef struct {
+    const tusb_desc_device_t *dev;      /*!< Pointer to device descriptor */
+    union {
+        const uint8_t *cfg;             /*!< Pointer to FullSpeed configuration descriptor when device one-speed only */
+        const uint8_t *fs_cfg;          /*!< Pointer to FullSpeed configuration descriptor when device support HighSpeed */
+    };
+#if (TUD_OPT_HIGH_SPEED)
+    const uint8_t *hs_cfg;              /*!< Pointer to HighSpeed configuration descriptor */
+    const tusb_desc_device_qualifier_t *qualifier;            /*!< Pointer to Qualifier descriptor */
+    uint8_t *other_speed;               /*!< Pointer for other speed configuration descriptor */
+#endif // TUD_OPT_HIGH_SPEED
+    const char *str[USB_STRING_DESCRIPTOR_ARRAY_SIZE];  /*!< Pointer to array of UTF-8 strings */
+    int str_count;                      /*!< Number of descriptors in str */
+} tinyusb_descriptor_config_t;
+
+static tinyusb_descriptor_config_t s_desc_cfg;
 
 // =============================================================================
 // CALLBACKS
@@ -28,7 +52,8 @@ static const char *s_str_descriptor[USB_STRING_DESCRIPTOR_ARRAY_SIZE]; // Array 
  */
 uint8_t const *tud_descriptor_device_cb(void)
 {
-    return (uint8_t const *)s_device_descriptor;
+    assert(s_desc_cfg.dev);
+    return (uint8_t const *)s_desc_cfg.dev;
 }
 
 /**
@@ -41,8 +66,53 @@ uint8_t const *tud_descriptor_device_cb(void)
 uint8_t const *tud_descriptor_configuration_cb(uint8_t index)
 {
     (void)index; // Unused, this driver supports only 1 configuration
-    return s_configuration_descriptor;
+    assert(s_desc_cfg.cfg);
+
+#if (TUD_OPT_HIGH_SPEED)
+    // HINT: cfg and fs_cfg are union, no need to assert(fs_cfg)
+    assert(s_desc_cfg.hs_cfg);
+    // Return configuration descriptor based on Host speed
+    return (TUSB_SPEED_HIGH == tud_speed_get())
+           ? s_desc_cfg.hs_cfg
+           : s_desc_cfg.fs_cfg;
+#else
+    return s_desc_cfg.cfg;
+#endif // TUD_OPT_HIGH_SPEED
 }
+
+#if (TUD_OPT_HIGH_SPEED)
+/**
+ * @brief Invoked when received GET DEVICE QUALIFIER DESCRIPTOR request
+ * Descriptor contents must exist long enough for transfer to complete
+ * If not highspeed capable stall this request
+ */
+uint8_t const *tud_descriptor_device_qualifier_cb(void)
+{
+    assert(s_desc_cfg.qualifier);
+    return (uint8_t const *)s_desc_cfg.qualifier;
+}
+
+/**
+ * @brief Invoked when received GET OTHER SPEED CONFIGURATION DESCRIPTOR request
+ * Descriptor contents must exist long enough for transfer to complete
+ * Configuration descriptor in the other speed e.g if high speed then this is for full speed and vice versa
+ */
+uint8_t const *tud_descriptor_other_speed_configuration_cb(uint8_t index)
+{
+    assert(s_desc_cfg.other_speed);
+
+    const uint8_t *other_speed = (TUSB_SPEED_HIGH == tud_speed_get())
+                                 ? s_desc_cfg.fs_cfg
+                                 : s_desc_cfg.hs_cfg;
+
+    memcpy(s_desc_cfg.other_speed,
+           other_speed,
+           ((tusb_desc_configuration_t *)other_speed)->wTotalLength);
+
+    ((tusb_desc_configuration_t *)s_desc_cfg.other_speed)->bDescriptorType = TUSB_DESC_OTHER_SPEED_CONFIG;
+    return s_desc_cfg.other_speed;
+}
+#endif // TUD_OPT_HIGH_SPEED
 
 /**
  * @brief Invoked when received GET STRING DESCRIPTOR request
@@ -54,11 +124,12 @@ uint8_t const *tud_descriptor_configuration_cb(uint8_t index)
 uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
 {
     (void) langid; // Unused, this driver supports only one language in string descriptors
+    assert(s_desc_cfg.str);
     uint8_t chr_count;
     static uint16_t _desc_str[MAX_DESC_BUF_SIZE];
 
     if (index == 0) {
-        memcpy(&_desc_str[1], s_str_descriptor[0], 2);
+        memcpy(&_desc_str[1], s_desc_cfg.str[0], 2);
         chr_count = 1;
     } else {
         if (index >= USB_STRING_DESCRIPTOR_ARRAY_SIZE) {
@@ -66,12 +137,12 @@ uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
             return NULL;
         }
 
-        if (s_str_descriptor[index] == NULL) {
+        if (s_desc_cfg.str[index] == NULL) {
             ESP_LOGW(TAG, "String index (%u) points to NULL, check your string descriptor", index);
             return NULL;
         }
 
-        const char *str = s_str_descriptor[index];
+        const char *str = s_desc_cfg.str[index];
         chr_count = strnlen(str, MAX_DESC_BUF_SIZE - 1); // Buffer len - header
 
         // Convert ASCII string into UTF-16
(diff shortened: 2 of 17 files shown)


device/esp_tinyusb/descriptors_control.c
```diff
@@ -41,9 +43,58 @@ uint8_t const *tud_descriptor_device_cb(void)
...
+uint8_t const *tud_descriptor_other_speed_configuration_cb(uint8_t index)
+{
+    assert(s_desc_cfg);
+    assert(s_desc_cfg->other_speed);
+
+    const uint8_t *other_speed = (TUSB_SPEED_HIGH == tud_speed_get())
+                                 ? s_desc_cfg->fs_cfg
+                                 : s_desc_cfg->hs_cfg;
+
+    memcpy(s_desc_cfg->other_speed,
```
tore-espressif, line 89: Why do we need to memcpy the descriptor? can't we just return to either HS or FS desc?

device/esp_tinyusb/descriptors_control.c
```diff
@@ -89,11 +142,109 @@ uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
...
+                         : descriptor_fs_cfg_kconfig;
+
+    // HS and FS cfg desc should be equal length
+    ESP_GOTO_ON_FALSE(((tusb_desc_configuration_t *)s_desc_cfg->hs_cfg)->wTotalLength ==
+                      ((tusb_desc_configuration_t *)s_desc_cfg->fs_cfg)->wTotalLength,
+                      ESP_ERR_INVALID_ARG, fail, TAG, "HS and FS Configuration descriptors must be same length");
+
+    // Qualifier Descriptor
+    if (NULL == config->qualifier_desc) {
+        ESP_LOGW(TAG, "Qualifier descriptor must be provided for HighSpeed device. Create one with current Device Descriptor.");
```
tore-espressif, line 197: Nitpick:
```suggestion
        ESP_LOGW(TAG, "Qualifier descriptor must be provided for HighSpeed device. Creating one from current Device Descriptor.");
```

roma-jam:
@tore-espressif Thanks for the review!
Will follow-up with all suggestions.

device/esp_tinyusb/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 1.4.4 (Unreleased)
 
+- esp_tinyusb: Configuration descriptor for HS/FS Hosts
```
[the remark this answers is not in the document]
roma-jam: Thanks, I just check the previous notes and made the same style:

    ## 1.4.2
    
    - MSC: Fix maximum files open
    - Add uninstall function
    
    ## 1.4.0
    
    - MSC: Fix integer overflows
    - CDC-ACM: Remove intermediate RX ringbuffer
    - CDC-ACM: Increase default FIFO size to 512 bytes
    - CDC-ACM: Fix Virtual File System binding


Do I need to change the tense in the old notes as well?

device/esp_tinyusb/include_private/descriptors_control.h
```diff
@@ -6,23 +6,46 @@
...
+typedef struct {
+    const tusb_desc_device_t *dev;      /*!< Pointer to device descriptor */
+    union {
+        const uint8_t *cfg;             /*!< Pointer to FS configuration descriptor when device one-speed only */
+        const uint8_t *hs_cfg;          /*!< Pointer to HS configuration descriptor when device one-speed only */
+    };
+#if (TUD_OPT_HIGH_SPEED)
+    const uint8_t *fs_cfg;              /*!< Pointer to FS configuration descriptor when device support HS */
+    tusb_desc_device_qualifier_t *qualifier;            /*!< Pointer to Qualifier descriptor */
+    bool qualifier_alloc_internally;    /*!< Qualifier descriptor was allocated  internally */
```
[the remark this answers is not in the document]
roma-jam: Because if it is not provided by user, it assembling just right there (Not to keep both descriptors and use the memory in case is has been presented by user). With copying data from Device Descriptor (that is why it is not `const`)
But in case changing the way of how descriptors is kept (static or dynamic) this part also could be changed and can be done simpler.

device/esp_tinyusb/descriptors_control.c
roma-jam: Because we need change the bDescriptorType to 0x07 (OTHER_SPEED_CONFIGURATION) in the buffer, which must exist long enough for transfer to complete.

If we return just the pointer - we return the descriptor with type 0x02 (CONFIGURATION) as a reply for GET_DESCRIPTOR request with descriptor type 0x07  (OTHER_SPEED_CONFIGURATION) in the wValue field.

device/esp_tinyusb/include_private/descriptors_control.h
```diff
@@ -6,23 +6,46 @@
...
+    uint8_t *other_speed;               /*!< Pointer for other speed configuration descriptor */
```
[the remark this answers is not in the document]
tore-espressif: Makes sense, I didn't realize the 0x07 difference!

device/esp_tinyusb/CHANGELOG.md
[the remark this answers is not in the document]
tore-espressif: Yes, please.

It's just a nitpick, but let's have the same format of releasenotes/changelog as in esp-idf :)

device/esp_tinyusb/CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 ## 1.4.4 (Unreleased)
```
tore-espressif, line 1:
```suggestion
## 1.4.4
```

tore-espressif:
@roma-jam Do you think we could increase default debug level to 1 = error? 
https://github.com/espressif/esp-usb/blob/master/device/esp_tinyusb/Kconfig#L4

``` c
// CFG_TUSB_DEBUG for debugging
// 0 : no debug
// 1 : print error
// 2 : print warning
// 3 : print info
```

roma-jam:
@tore-espressif Sure. 

Otherwise, shouldn't we turn off the debug at all in the release and implement asserts when we need to use debug error message for that? 
Anyway, seems that device won't work as expected after that error, so maybe it is better to keep the debug output for debug purpose? WDYT?

device/esp_tinyusb/include_private/descriptors_control.h
```diff
@@ -34,6 +36,12 @@ void tinyusb_set_descriptor(const tusb_desc_device_t *dev_desc, const char **str
  */
 void tinyusb_set_str_descriptor(const char *str, int str_idx);
 
+/**
+ * @brief Free memory allocated during tinyusb_set_descriptors
+ *
+ */
+void tinyusb_free_desctiptors(void);
```
tore-espressif, line 43: Should this function be called during tinyusb_uninstall?

device/esp_tinyusb/include/tinyusb.h
```diff
@@ -31,7 +31,20 @@ typedef struct {
     const char **string_descriptor;            /*!< Pointer to array of string descriptors. If set to NULL, TinyUSB device will use a default string descriptors whose values are set in Kconfig */
     int string_descriptor_count;               /*!< Number of descriptors in above array */
     bool external_phy;                         /*!< Should USB use an external PHY */
-    const uint8_t *configuration_descriptor;   /*!< Pointer to a configuration descriptor. If set to NULL, TinyUSB device will use a default configuration descriptor whose values are set in Kconfig */
+    union {
+        struct {
+            const uint8_t *configuration_descriptor; /*!< Pointer to a configuration descriptor. If set to NULL, TinyUSB device will use a default configuration descriptor whose values are set in Kconfig */
+            const uint8_t *reserved1;                /*!< For backward compatibility */
+            const uint8_t *reserved2;                /*!< For backward compatibility */
```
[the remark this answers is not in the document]
tore-espressif: I can still see some `reserved` fields. Why are they there?

tore-espressif (commented):
Thanks @roma-jam , looks better now!

device/esp_tinyusb/test/local/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.22)
+project(libusb_test
```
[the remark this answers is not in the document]
roma-jam: Yes, but ATM I stuck with the problem, that I can't open a device from python. I have tried to solve it and then decided to write it in C, because it was faster and it worked. 

Ok, will do that, but I need some time to understand why I can't open device from python.

device/esp_tinyusb/include/tinyusb.h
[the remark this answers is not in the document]
roma-jam: Briefly: to exclude the undefined behavior.

In general, the idea to make the union of two structs with one and two pointers means, that during the configuration and structure initialization (`const tinyusb_config_t tusb_cfg = { ... }`)  we use old `.configuration_descriptor` member and we can use new member for `.fs_cfg_desc`. 

Problem appears when compiler creates the structure and fill the memory. To be more specific - in the way how he solves the two-structs union initialization. If compiler makes it one after another, that means, that is we are initialize not all members of structure, then the memory will be initialized with the last one. 

Here you may find the simple example-code (to verify it in OnlineGDB, for example): 
```
static uint8_t const config_desc[] = {
    'a', 'b', 'c',
};

static uint8_t const fs_config_desc[] = {
    'w', 's', 'x'
};

typedef struct {
    union { 
        struct {
            const uint8_t *ptr1_1; 
            const uint8_t *ptr1_2;
        };
        struct {
            const uint8_t *ptr2_1;   
            const uint8_t *ptr2_2;              
        };
    };
} descriptor_t; 

void print_desc(const uint8_t * ptr) 
{
    if (ptr == NULL) {
        printf("NULL\r\n");
        return;
    }
    for (int i = 0; i < 3; i++) {
        printf("%c ", ptr[i]);
    }
    printf("\n\r");
}

int main()
{
    const descriptor_t d = {
        .ptr1_1 = config_desc,
        // .ptr1_2 = config_desc,
        // .ptr2_1 = fs_config_desc,
        .ptr2_2 = fs_config_desc
    };
    
    printf("size %ld\r\n", sizeof(descriptor_t));
    
    for(int i = 0; i < sizeof(descriptor_t); i++) {
        printf("%02X ", *((uint8_t*)(&d) + i));
    }
    printf("\r\n");
    
    print_desc(d.ptr1_1);
    print_desc(d.ptr1_2);
    print_desc(d.ptr2_1);
    print_desc(d.ptr2_2);

    return 0;
}
```

The result of the code: 
```
size 16
00 00 00 00 00 00 00 00 0A 40 A4 CB 52 56 00 00 
NULL
w s x 
NULL
w s x 
```

We can see, that the structure contains the data of the initialization values for second structure in union (`.ptr2_2 = fs_config_desc` and `.ptr2_1` is not defined ).  

For RISCV it works a bit differentle and when we have the `.reserved` field, the initialization of `.fs_cfg_desc` doesn't rewrite the `.configuration_descriptor` member with `.hs_cfg_desc` (which is not defined, as we don't change the example code).

So, either we need to force to change the example and use always new name for `.hs_cfg_desc`, or add `.reserved` field. 
I decided to do the second option, so we can postpone with updating the example code.

device/esp_tinyusb/test/local/CMakeLists.txt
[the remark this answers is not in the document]
tore-espressif: I see. I'd like to merge this soon so we can release esp_tinyusb 1.4.4. Please rebase on top of #17 .

In case you can't make python work, we can merge it without it, but please follow-up with test refactor ASAP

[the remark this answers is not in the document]
roma-jam: @tore-espressif 
Rebase done.
I also have added the "hot_disconnect" manual test, to verify the problem with halting the device while disconnect happens during ongoing transfer.
Firmware added, but software (pytest_usb_device_hot_disconnect.py) part is not. I tried to implement the host part (opening port, receive data, re-opening port after exception while device is gone), but it is not so trivial. I will check how it is done in esp-idf monitor and try to come up with something (1). 

To run esp_tinyusb/device tests in pytest correctly, tinyusb_uninstall function need to be added to the tests, so we could launch them one after another (2). 

And rewrite device qualifier and other speed descriptor test to python. (3)

I will add them to the plan, one thing is left before it can be merged - make a decision about the LOG level in tinyusb.

device/esp_tinyusb/include/tinyusb.h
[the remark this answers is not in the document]
tore-espressif: @roma-jam you **cannot initialize multiple members of a union!**

Unfortunately, GCC will not warn you about this. It is the programmer's responsibility. Clang based compilers will give you a warning in C:
```
warning: initializer overrides prior initialization of this subobject [-Winitializer-overrides]
<source>:41:9: note: previous initialization is here
        .ptr1_1 = config_desc,
        ^~~~~~~~~~~~~~~~~~~~~
```

Just for completeness, you won't be able to compile the same code with C++ compilers, which are more strict.
Most of them (GCC based) will get you this error:
```
error: too many initializers for 'const descriptor_t'
```
Which is correct because you **cannot initialize multiple members of a union!**
Some other compilers (Clang based) give a better error message:
```
error: initializer overrides prior initialization of this subobject [-Winitializer-overrides]
<source>:40:9: note: previous initialization is here
        .ptr1_1 = config_desc,|
```

tore-espressif (changes requested):

device/esp_tinyusb/include/tinyusb.h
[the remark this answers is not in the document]
roma-jam: Thanks for the detailed explanation!

tore-espressif (commented):
Thank you @roma-jam LGTM.

@Dazza0 PTAL too

device/esp_tinyusb/descriptors_control.c
```diff
@@ -89,11 +160,94 @@ uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
...
+        // Default configuration descriptor is provided only for CDC, MSC and NCM classes
+#if (CFG_TUD_HID > 0 || CFG_TUD_MIDI > 0 || CFG_TUD_CUSTOM_CLASS > 0 || CFG_TUD_ECM_RNDIS > 0 || CFG_TUD_DFU > 0 || CFG_TUD_DFU_RUNTIME > 0 || CFG_TUD_BTH > 0)
+        ESP_GOTO_ON_FALSE(config->hs_cfg_desc, ESP_ERR_INVALID_ARG, fail, TAG, "Configuration descriptor must be provided for this device");
+#else
+        ESP_LOGW(TAG, "The device's HS configuration descriptor is not provided by user, using default.");
+#endif
+    }
+    s_desc_cfg.cfg = config->hs_cfg_desc
+                     ? config->hs_cfg_desc
+                     : descriptor_cfg_kconfig;
```
Dazza0, line 190: Nitpick: Likewise

```suggestion
        s_desc_cfg.cfg = descriptor_cfg_kconfig;
    } else {
        s_desc_cfg.cfg = config->hs_cfg_desc;
    }
```

device/esp_tinyusb/descriptors_control.c
```diff
@@ -89,11 +160,94 @@ uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
...
+        // Default configuration descriptor is provided only for CDC, MSC and NCM classes
+#if (CFG_TUD_HID > 0 || CFG_TUD_MIDI > 0 || CFG_TUD_CUSTOM_CLASS > 0 || CFG_TUD_ECM_RNDIS > 0 || CFG_TUD_DFU > 0 || CFG_TUD_DFU_RUNTIME > 0 || CFG_TUD_BTH > 0)
+        ESP_GOTO_ON_FALSE(config->fs_cfg_desc, ESP_ERR_INVALID_ARG, fail, TAG, "Configuration descriptor must be provided for this device");
+#else
+        ESP_LOGW(TAG, "The device's FS configuration descriptor is not provided by user, using default.");
+#endif
+    }
+    s_desc_cfg.fs_cfg = config->fs_cfg_desc
+                        ? config->fs_cfg_desc
+                        : descriptor_fs_cfg_kconfig;
```
Dazza0, line 203: Nitpick: Likewise
```suggestion
        s_desc_cfg.fs_cfg = descriptor_fs_cfg_kconfig;
    } else {
        s_desc_cfg.fs_cfg = config->fs_cfg_desc;
    }
```

device/esp_tinyusb/descriptors_control.c
```diff
@@ -89,11 +160,94 @@ uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
...
+        ESP_LOGW(TAG, "The device's qulifier descriptor is not provided by user, using default.");
+        s_desc_cfg.qualifier = &descriptor_qualifier_kconfig;
+    } else {
+        s_desc_cfg.qualifier = config->qualifier_descriptor;
+    }
+
+    // Other Speed buffer allocate
+    s_desc_cfg.other_speed = calloc(1, ((tusb_desc_configuration_t *)s_desc_cfg.hs_cfg)->wTotalLength);
+    ESP_GOTO_ON_FALSE(s_desc_cfg.other_speed, ESP_ERR_NO_MEM, fail, TAG, "Other speed memory allocation error");
+#else
```
Dazza0, line 223: Nitpick: For large `#if`/`#else`/`#endif` blocks, suggesting adding a comment of the `#if` condition to make the code more readable.

```suggestion
#else // TUD_OPT_HIGH_SPEED
```

device/esp_tinyusb/descriptors_control.c
```diff
@@ -89,11 +160,94 @@ uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
...
+        // Default configuration descriptor is provided only for CDC, MSC and NCM classes
+#if (CFG_TUD_HID > 0 || CFG_TUD_MIDI > 0 || CFG_TUD_CUSTOM_CLASS > 0 || CFG_TUD_ECM_RNDIS > 0 || CFG_TUD_DFU > 0 || CFG_TUD_DFU_RUNTIME > 0 || CFG_TUD_BTH > 0)
+        ESP_GOTO_ON_FALSE(config->configuration_descriptor, ESP_ERR_INVALID_ARG, fail, TAG, "Configuration descriptor must be provided for this device");
+#else
+        ESP_LOGW(TAG, "The device's configuration descriptor is not provided by user, using default.");
+#endif
+    }
+    s_desc_cfg.cfg = config->configuration_descriptor
+                     ? config->configuration_descriptor
+                     : descriptor_cfg_kconfig;
```
Dazza0, line 234: Likewise
```suggestion
        s_desc_cfg.cfg = descriptor_cfg_kconfig;
    } else {
        s_desc_cfg.cfg = config->configuration_descriptor;
    }
```

device/esp_tinyusb/descriptors_control.c
```diff
@@ -89,11 +160,94 @@ uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
...
+    assert(config);
+    const char **pstr_desc;
+    // Parse configuration and save descriptors's pointer
+    // Select Device Descriptor
+    if (NULL == config->device_descriptor) {
+        ESP_LOGW(TAG, "The device's device descriptor is not provided by user, using default.");
+    }
+    s_desc_cfg.dev = config->device_descriptor
+                     ? config->device_descriptor
+                     : &descriptor_dev_kconfig;
```
Dazza0, line 175: Nitpick: IMO this is cleaner. Avoids the double conditional statements.

```suggestion
        s_desc_cfg.dev = &descriptor_dev_kconfig;
    } else {
        s_desc_cfg.dev = config->device_descriptor;
    }
```

device/esp_tinyusb/CHANGELOG.md
```diff
@@ -1,43 +1,45 @@
...
+## 1.4.4
+
+- esp_tinyusb: Added possibility to set FS configuration descriptor for HS devices
+- CDC-ACM: Removed MIN() definition if already defined
+- MSC: Fixed EP size selecting in default configuration descriptor
 
-- CDC-ACM: Remove MIN() definition if already defined
-- MSC: Set EP size in configuration descriptor based on speed
 
 ## 1.4.3
```
Dazza0, line 8: Nitpick: Extra line

```suggestion
## 1.4.3
```

device/esp_tinyusb/include_private/usb_descriptors.h
```diff
@@ -30,12 +40,23 @@ extern const char *descriptor_str_kconfig[];
...
+ * The user can provide his own HighSpeed or FullSpeed configuration descriptor during tinyusb_driver_install() call
  */
 extern const uint8_t descriptor_cfg_kconfig[];
 
+#if (TUD_OPT_HIGH_SPEED)
+/**
+ * @brief FullSpeed Configuration descriptor generated from Kconfig
+ *
+ * This descriptor is used by default.
+ * The user can provide his own descriptor during tinyusb_driver_install() call
```
Dazza0, line 55:
```suggestion
 * The user can provide their own descriptor via tinyusb_driver_install()
```

device/esp_tinyusb/usb_descriptors.c
```diff
@@ -62,6 +62,30 @@ const tusb_desc_device_t descriptor_dev_kconfig = {
     .bNumConfigurations = 0x01
 };
 
+#if (TUD_OPT_HIGH_SPEED)
+const tusb_desc_device_qualifier_t descriptor_qualifier_kconfig = {
```
Dazza0, line 66: @roma-jam @tore-espressif Given that these are private, could we also rename the `kconfig` descriptors to default? I get that some of their values are set via Kconfig. But feature wise, they act as the default descriptors if none are provided by the user.

```suggestion
const tusb_desc_device_qualifier_t descriptor_qualifier_default = {
```

Likewise with the other `..._kconfig` descriptors in this header as well.

Dazza0 (approved):
@roma-jam Left a bunch of nitpicks, but changes LGTM overall

tore-espressif (approved):

device/esp_tinyusb/usb_descriptors.c
roma-jam: I don't mind, I like `_default` more than `_kconfig`. 
Probably, for Device Descriptor `_kconfig` makes sense, because we can configure the VID/PID with KConfig.
But anyway it is the default descriptor which is used when user doesn't provide any other.

roma-jam: @tore-espressif may I briefly ask you about your opinion? 
Rename `descriptor_name_kconfig` -> `descriptor_name_default`?

device/esp_tinyusb/test_app/main/test_tud_cb.c
```diff
@@ -0,0 +1,28 @@
...
+
+#include <stdio.h>
+#include <string.h>
+#include "tinyusb.h"
+#include "tusb_tasks.h"
+#include "test_bvalid_sig.h"
+#include "test_descriptors_config.h"
+
+// Invoked when device is mounted
+void tud_mount_cb(void)
```
roma-jam, line 15: @tore-espressif  I have added the test to verify all possible configuration options. 
The base flag of successful mounting - `tud_mount_cb`. 
But I need that callback in both tests (`bvalid_signal` and `descriptors_config`), so I decided to make the general callback and then to call callbacks within the specific test from it. 

I am not 100% sure that this is the best way, maybe you can suggest something better.

device/esp_tinyusb/usb_descriptors.c
tore-espressif: I agree with renaming to `_default`

device/esp_tinyusb/test_app/main/test_tud_cb.c
tore-espressif: Now the 2 tests are interfering with one another: running the 'descriptor' test case will modify value of `dev_mounted/umounted` inside the 'bvalid' testcases.

If you plan to merge it like this, please track the follow-up somewhere, so we can fix soon

PR #13, opened 2024-02-07
fix(cdc_acm): Fix closing of already closed device

tore-espressif:
This bug was found out during esp-serial-flasher CDC operation.

In case the user does not track status of the CDC device and attempts to close the device from multiple places/threads the driver would crash

cc @DNedic


roma-jam (approved):
For some reason this PR contain the same changes as in feature/cdc_func_deacriptors despite the fact that the target branch is the same :shrug: 
Otherwise, changes about closing already closed device LGTM.

tore-espressif:
@roma-jam Thank you for the review!

> For some reason this PR contain the same changes as in feature/cdc_func_deacriptors

I rebased [feature/cdc_func_descriptors](https://github.com/espressif/esp-usb/tree/feature/cdc_func_descriptors) on master so that is why you see 2 commits. I'll fix before merging

PR #15, opened 2024-02-12
CI: Add copyright license check

tore-espressif:
Added copyright header check to pre-commit

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,6 +11,12 @@ repos:
     - id: codespell
       args: [-w, "--ignore-words=codespell-ignore-list"]
 
+-   repo: https://github.com/espressif/check-copyright/
+    rev: v1.0.3
+    hooks:
+      - id: check-copyright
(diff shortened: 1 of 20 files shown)


roma-jam (approved):
LGTM.
Should we added the changes also to a git workflow?

peter-marcisovsky (approved):
I was checking whether you need to update the years on the license if you are changing the license itself, not the actual code. 

Found you it's ok the way as it is now. But, correct me if I am wrong.
Otherwise, LGTM.

tore-espressif:
Thanks for the reviews!

@roma-jam  what do you mean by 

> Should we added the changes also to a git workflow?

If you have pre-commit installed the check will run automatically, so there is no need to do anything extra

roma-jam:
@tore-espressif The same as you did here: https://github.com/espressif/esp-usb/pull/12?

tore-espressif:
#12 added pre-commit run to CI. 

So this change will be automatically reflected in your local repo and also checked in CI

PR #27, opened 2024-03-22
feat: add usb host uac driver and examples (IEC-99)

leeebo:
The PR aims to add UAC Host Driver (Based on ESP-IDF USB Host Driver), featuring:
- Compatibility with UAC 1.0 specification
- Independent IN/OUT streaming control
- Independent Volume/Mute control
- Multiple devices/instances for future HUB support

Checklist:

- [x] (No mock test, can only be test locally) Speaker and Mic works as normal
~~- [x] example: Play MP3 as normal~~ I have removed the examples to [esp-iot-solution/examples](https://gitlab.espressif.cn:6688/ae_group/esp-iot-solution/-/merge_requests/974), reducing the work of review here.
~~- [ ] example: Record voice to SDcard~~

- [x] thread-safe protect (ringbuffer safe delete, critical section protect)
- [x] refactor open/close device as @roma-jam suggest
- [x]  Support Parse USB Audio device with MIXER_UNIT [see](https://github.com/espressif/esp-usb/pull/27#issuecomment-2082126829)

uac_host_refactor2


leeebo:
@tore-espressif @roma-jam @Dazza0 PTAL!

roma-jam:
Hey @leeebo, 
Wow, a lot of code, great job! 
I briefly checked the logic, and it seems that it is the same as we have right now in hid_driver v1.0.2 (correct me if I am wrong).

The main problem there - is that if device has at least one interface, driver opens the device with `usb_host_device_open()` and hold the device opened, even if the interface is not claimed (opened) by user. This happened during the `client_event_cb` which we should fulfill as fast as we can.

In that case, it is better to "touch" the new device, checking the uac interface presence and add it to the list.
I made the changes in hid driver here (but they are not ready to be merge yet): https://github.com/espressif/idf-extra-components/pull/214

But I explained the idea in https://github.com/espressif/idf-extra-components/pull/214#issuecomment-1761262534 

Please, read the document and share your opinion.

leeebo:
@roma-jam Yes, you are right, the device handling logic and data structure is same as HID v1.0.2. Thanks for yor advice, I will check it.

peter-marcisovsky:
@leeebo regarding the test app: 

I took a look at the failing CI logs, could you please try renaming the `pytest_usb_host_hid.py` to `pytest_usb_host_uac.py` in your `test_app` folder? So we will have passing CI for this PR.

leeebo:
> @leeebo regarding the test app:
> 
> I took a look at the failing CI logs, could you please try renaming the `pytest_usb_host_hid.py` to `pytest_usb_host_uac.py` in your `test_app` folder? So we will have passing CI for this PR.

Hi @peter-marcisovsky , as we currently have no UAC 1.0 device for the mock test (tinyusb supports UAC2.0 only) , I just test it using my local UAC device, any advice for the test?

peter-marcisovsky:
@leeebo sorry I was not aware that the tinyusb supports only UAC2.0. 

In this case, please enable at least build in CI by adding a path to your new component to `.idf_build_apps.toml` file in the root dir.

host/class/uac/usb_host_uac/uac_descriptors.c
```diff
@@ -0,0 +1,313 @@
...
+    }
+    return;
+unknown:
+    print_unknown_desc(header);
+}
+
+// Print the configuration descriptor and all its sub-descriptors with the given class-specific callback
+// The subclass and protocol are passed to the class_specific_cb to allow it to interpret the descriptors more accurately
+// This function better be added to usb_helpers.c
+static void usb_print_config_descriptor_with_context(const usb_config_desc_t *cfg_desc, print_class_descriptor_with_context_cb class_specific_cb)
```
leeebo, line 272: We better to add the helper function to IDF, the Class driver need to know the interface context to parse the CS interface/endpoint which belong to it.

host/class/uac/usb_host_uac/test_app/sdkconfig.defaults
```diff
@@ -0,0 +1,8 @@
+# This file was generated using idf.py save-defconfig. It can be edited manually.
+# Espressif IoT Development Framework (ESP-IDF) Project Minimal Configuration
+#
+CONFIG_ESPTOOLPY_FLASHSIZE_4MB=y
+CONFIG_PARTITION_TABLE_CUSTOM=y
+CONFIG_FREERTOS_HZ=1000
+CONFIG_USB_HOST_CONTROL_TRANSFER_MAX_SIZE=2048
+CONFIG_USB_HOST_HW_BUFFER_BIAS_PERIODIC_OUT=y
```
leeebo, line 8: In addition to providing three FIFO allocation logic, ESP-IDF need more flexible support for user configuration of FIFO sizes, so that we can provide more flexible device support.

I have a USB Headset with IN MPS 192, OUT MPS 384, there is no FIFO configuration that allows two interfaces to be used simultaneously.

leeebo:
Another topic is whether we can adjust ep's MPS before declaring the interface, which, although not standard, makes it easier to provide workaround when the FIFO size is limited.

In this Headset case, the speaker supports sampling rate 96KHz 16bit 2ch at maximum, so the OUT MPS is 96*2*2 = 384. It is used when the user selects 96KHz, and will not be touched in most time. It is sufficient to declare an interface with MPS=192~193 when selecting 48KHz.

image

host/class/uac/usb_host_uac/include/usb/uac_host.h
```diff
@@ -0,0 +1,374 @@
...
+ *
+ * The maximum possible number of characters in an embedded string is device specific.
+ * For USB devices, the maximum string length is 126 wide characters (not including the terminating NULL character).
+ * This is a length, which is available to upper level application during getting information
+ * of UAC Device with 'uac_host_get_device_info' call.
+ *
+ * To decrease memory usage 32 wide characters (64 bytes per every string) is used.
+*/
+#define UAC_STR_DESC_MAX_LENGTH              (32)
+#define FLAG_STREAM_SUSPEND_AFTER_START      (1 << 0)      /*!< suspend stream after start, then call uac_host_device_resume to resume */
```
tore-espressif, line 29: Could we add a better description about where this flag is supposed to be used?

host/class/uac/usb_host_uac/include/usb/uac_host.h
```diff
@@ -0,0 +1,374 @@
...
+/**
+ * @brief USB UAC driver event callback.
+ *
+ * @param[in] addr        USB Address of connected UAC device
+ * @param[in] iface_num   UAC Interface Number
+ * @param[in] event       UAC driver event
+ * @param[in] arg         User argument from UAC driver configuration structure
+*/
+typedef void (*uac_host_driver_event_cb_t)(uint8_t addr, uint8_t iface_num,
+        const uac_host_driver_event_t event, void *arg);
```
tore-espressif, line 62: For future extension of this API, it might be better to prepare a field for event specific data, eg.
``` c
struct {
    uac_host_device_event_t event_type;
    union {
        // For future extensions
    } event_data;
}
```

host/class/uac/usb_host_uac/include/usb/uac_host.h
```diff
@@ -0,0 +1,374 @@
...
+*/
+typedef struct {
+    uint8_t channels;                                    /*!< Audio channel number */
+    uint8_t bit_resolution;                              /*!< Audio bit resolution */
+    uint32_t sample_freq;                                /*!< Audio sample resolution */
+    uint16_t flags;                                      /*!< Control flags */
+} uac_host_stream_config_t;
+
+/**
+ * @brief UVC control type
```
tore-espressif, line 166: typo
```suggestion
 * @brief UAC control type
```

host/class/uac/usb_host_uac/include/usb/uac_host.h
```diff
@@ -0,0 +1,374 @@
...
+ * @param[in] uac_dev_handle  UAC device handle
+ * @param[in] type            Control type
+ * @param[in] value           Control value
+ * @return esp_err_t
+ *  - ESP_OK on success
+ *  - ESP_ERR_INVALID_STATE if UAC device is not ready or active
+ *  - ESP_ERR_INVALID_ARG if the device handle is invalid
+ *  - ESP_ERR_NOT_SUPPORTED if the control type is not supported
+ */
+esp_err_t uac_host_device_control(uac_host_device_handle_t uac_dev_handle, uac_host_ctrl_t type, void *value);
```
tore-espressif, line 370: I would prefer having separate function for each of the control options. It would help us getting rid of the `void*` pointer

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -0,0 +1,2114 @@
...
+            // we need to return invalid state to safely exit the write function
+            if (UAC_INTERFACE_STATE_ACTIVE != iface->state) {
+                ret = ESP_ERR_INVALID_STATE;
+                goto exit_critical;
+            }
+            iface->xfer_list[i] = iface->free_xfer_list[i];
+            iface->free_xfer_list[i] = NULL;
+            iface->xfer_list[i]->status = USB_TRANSFER_STATUS_COMPLETED;
+            UAC_EXIT_CRITICAL();
+            stream_tx_xfer_done(iface->xfer_list[i]);
```
tore-espressif, line 2070: Do I understand it right that here we start the transfer by calling transfer done callback?
If yes, I suggest modifying to logic as this seems to be counter intuitive. Or at least add a comment about what and why are we doing here

tore-espressif (commented):
@leeebo I'm sorry for the late review.

This is 1st round review focused mostly on API + some nitpicks.

I'll continue with implementation review soon

host/class/uac/usb_host_uac/include/usb/uac_host.h
```diff
@@ -0,0 +1,374 @@
...
+
+/**
+ * @brief  USB UAC host class descriptor print callback
+ *
+ * @param[in] desc  Pointer to the USB configuration descriptor
+ * @param[in] subclass  Subclass of the UAC device
+ * @param[in] protocol  Protocol of the UAC device
+ *
+ */
+typedef void (*print_class_descriptor_with_context_cb)(const usb_standard_desc_t *,
```
peter-marcisovsky, line 82:
```suggestion
typedef void (*print_class_descriptor_with_context_cb)(const usb_standard_desc_t *desc,
```

host/class/uac/usb_host_uac/include/usb/uac_host.h
```diff
@@ -0,0 +1,374 @@
...
+ *  - ESP_ERR_INVALID_STATE if UAC driver is already installed
+ *  - ESP_ERR_INVALID_ARG if the configuration is invalid
+ *  - ESP_ERR_NO_MEM if memory allocation failed
+ *
+*/
+esp_err_t uac_host_install(const uac_host_driver_config_t *config);
+
+/**
+ * @brief Uninstall USB Host UAC Class driver
+ *
```
peter-marcisovsky, line 194: Can you add some note about closing device before calling this function? Something like:

```suggestion
 *
 * note This function must be called after a device was closed
```

host/class/uac/usb_host_uac/include/usb/uac.h
```diff
@@ -0,0 +1,480 @@
...
+typedef enum {
+    UAC_FORMAT_TYPE_UNDEFINED                         = 0x0000,
+    UAC_FORMAT_TYPE_I                                 = 0x0001,
+    UAC_FORMAT_TYPE_II                                = 0x0002,
+    UAC_FORMAT_TYPE_III                               = 0x0003
+} uac_format_type_t;
+
+/**
+ * @brief Audio Class-Specific Interface Descriptor
+ *
```
peter-marcisovsky, line 352: Does this `typedef` have also `see` option referring to some specific table in the docs?

host/class/uac/usb_host_uac/include/usb/uac.h
```diff
@@ -0,0 +1,480 @@
...
+typedef struct {
+    uint8_t bLength;
+    uint8_t bDescriptorType;
+    uint8_t bDescriptorSubtype;
+    uint8_t bTerminalLink;
+    uint8_t bDelay;
+    uint16_t wFormatTag;
+} __attribute__((packed)) uac_as_general_desc_t;
+
+#define UAC_FREQ_NUM_MAX                  3
```
peter-marcisovsky, line 439: Can you please add a comment here? Explaining why `UAC_FREQ_NUM_MAX` equals to 3.

host/class/uac/usb_host_uac/uac_descriptors.c
leeebo: The cb type is different, here defined a new callback function type with  `uint8_t subclass` as context

```
typedef void (*print_class_descriptor_with_context_cb)(const usb_standard_desc_t *, uint8_t subclass, uint8_t protocol);
```

peter-marcisovsky (commented):
Left some minor comments and typos fixes.

host/class/uac/usb_host_uac/uac_descriptors.c
```diff
@@ -0,0 +1,313 @@
...
+#include <inttypes.h>
+#include "usb/usb_helpers.h"
+#include "usb/usb_types_ch9.h"
+#include "esp_check.h"
+#include "usb/usb_host.h"
+#include "usb/uac_host.h"
+
+// ----------------------------------------------- Descriptor Printing -------------------------------------------------
+
+static void print_ep_desc(const usb_ep_desc_t *ep_desc)
```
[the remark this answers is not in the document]
leeebo: currently copy-pasted here for the implement of `usb_print_config_descriptor_with_context`

host/class/uac/usb_host_uac/include/usb/uac.h
```diff
@@ -0,0 +1,480 @@
...
+
+/**
+ * @brief 3D Stereo Extender Processing Unit Control Selectors
+ *
+ * @see Table A.10.3.3 of audio10.pdf
+ */
+typedef enum {
+    UAC_3DSE_CONTROL_UNDEFINED                        = 0x00,
+    UAC_3DSE_ENABLE_CONTROL                           = 0x01,
+    UAC_3DSE_SPACIOUSNESS_CONTROL                     = 0x02
```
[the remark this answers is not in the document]
leeebo: Oh, thank! it's so strange... I didn't notice that it was discontinuous

host/class/uac/usb_host_uac/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+idf_component_register( SRCS "uac_descriptors.c" "uac_host.c" "uac_host.c"
```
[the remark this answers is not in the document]
leeebo: fixed, thanks

host/class/uac/usb_host_uac/include/usb/uac.h
leeebo: There is no limitations on the number of `samples frequence` uac stream interface can offer. But for UAC1.0 devices, the number is usually not more than three. The best solution is dynamically malloc space for the supported frequencies, but this complicates the API, requiring the user to read the length first and then read the exact value.

Any suggestions? how about make it configurable in Kconfig?

host/class/uac/usb_host_uac/include/usb/uac.h
leeebo: No refer table, It's the common part of Class-Specific Descriptor, used for iterate over different types of descriptors

host/class/uac/usb_host_uac/include/usb/uac.h
```diff
@@ -0,0 +1,480 @@
+/*
+ * SPDX-FileCopyrightText: 2023-2024 Espressif Systems (Shanghai) CO LTD
```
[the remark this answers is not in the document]
leeebo: ok

host/class/uac/usb_host_uac/include/usb/uac.h
```diff
@@ -0,0 +1,480 @@
...
+    UAC_TYPE_I_ALAW                                   = 0x0004,
+    UAC_TYPE_I_MULAW                                  = 0x0005
+} uac_type_i_format_t;
+
+/**
+ * @brief Format Type Codes
+ *
+ */
+typedef enum {
+    UAC_FORMAT_TYPE_UNDEFINED                         = 0x0000,
```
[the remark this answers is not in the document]
leeebo: thanks, accept

host/class/uac/usb_host_uac/include/usb/uac.h
leeebo: let's change the brief to `Audio Class-Specific Interface Descriptor Common Header`

host/class/uac/usb_host_uac/include/usb/uac.h
```diff
@@ -0,0 +1,480 @@
...
+ * @see Table 4-7 of audio10.pdf
+ */
+typedef struct {
+    uint8_t bLength;
+    uint8_t bDescriptorType;
+    uint8_t bDescriptorSubtype;
+    uint8_t bUnitID;
+    uint8_t bSourceID;
+    uint8_t bControlSize;
+    uint8_t bmaControls[3 * 2];
```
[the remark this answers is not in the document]
leeebo: It's quite tricky descriptor, I can't define it well in a struct. `bmaControls` size is variable, it depends on the audio `bNrChannels` and `bControlSize`, here we assume (for most device) the audio device is dual-channel (3=2+1 (chanel 0)), each channel's `bControlSize` is 2 bytes. 

But it's should no problem for any other case. Because the real descriptor is parsed as it is

host/class/uac/usb_host_uac/include/usb/uac.h
```diff
@@ -0,0 +1,480 @@
...
+ */
+typedef enum {
+    UAC_TE_CONTROL_UNDEFINED                          = 0x00,
+    UAC_COPY_PROTECT_CONTROL                          = 0x01
+} uac_te_control_selector_t;
+
+/**
+ * @brief Feature Unit Control Selectors
+ *
+ * @see Table A.10.2 of audio10.pdf
```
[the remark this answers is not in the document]
leeebo: fixed, thanks

host/class/uac/usb_host_uac/include/usb/uac_host.h
```diff
@@ -0,0 +1,374 @@
...
+typedef enum {
+    UAC_HOST_DRIVER_EVENT_RX_CONNECTED = 0x00,           /*!< UAC RX Device has been found in connected USB device */
+    UAC_HOST_DRIVER_EVENT_TX_CONNECTED,                  /*!< UAC TX Device has been found in connected USB device */
+} uac_host_driver_event_t;
+
+/**
+ * @brief USB UAC Device (Interface) event id
+*/
+typedef enum {
+    UAC_HOST_DEVICE_EVENT_RX_DONE = 0x00,                /*!< Data in the rx buffer is higher than the threshold */
```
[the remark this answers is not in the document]
leeebo: Yes... To reduce the notifications, users can receive RX_DONE notifications only when the received data count reaches the threshold.

How about `/*!< RX Done: the receive buffer data exceeds the threshold */`

host/class/uac/usb_host_uac/uac_host.c
leeebo: yes, call the function is to simply keep all tranfer operate in one function.

let's split it into two functions to make it clear

host/class/uac/usb_host_uac/include/usb/uac_host.h
leeebo: ok, agree

host/class/uac/usb_host_uac/include/usb/uac_host.h
leeebo: ok, added

host/class/uac/usb_host_uac/include/usb/uac_host.h
leeebo: will consider it after the second round of review

tore-espressif (commented):
@leeebo Output path works great with my UAC 1.0 device!

~~Is this version supposed to work in input devices (microphones) too?~~

**EDIT:** I run a quick test with a microphone and it looks OK too

roma-jam:
Hi @leeebo, 
I have only two UAC devices I could use.
One of them fails with: 

```
I (3253) UAC_TEST: USB Host installed
I (3253) UAC_TEST: UAC Class Driver installed
E (128523) HCD DWC: EP MPS (200) exceeds supported limit (128)
E (128523) uac-host: uac_host_interface_claim_and_prepare_transfer(897): Unable to claim Interface
E (128528) uac-host: uac_host_device_start(1930): Unable to claim Interface
```
Which seems OK, as this device has two audio interfaces, where mps for EP is 200 B.

Another one fails with: 
```
 (13845) UAC_TEST: USB Host installed
I (13845) UAC_TEST: UAC Class Driver installed
W (19301) uac-host: Feature Unit ID 2 not corresponding to the input terminal
W (19301) uac-host: Feature Unit ID 7 not corresponding to the input terminal

assert failed: block_trim_free tlsf.c:496 (block_is_free(block) && "block must be free")
HINT: CORRUPT HEAP: heap metadata corrupted resulting in TLSF malfunction.
Make sure you are not making out of bound writing on the memory you allocate in your application.
Make sure you are not writing on freed memory.
For more information run 'idf.py docs -sp api-reference/system/heap_debug.html'.
```

Anyway, it seems that I am not sure how to verify it on a hardware. 
I am launching test_app, test 5 and then connect the device to the port. Am I doing something wrong?

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -0,0 +1,2141 @@
...
+/**
+ * @brief UAC IN Transfer complete callback
+ *
+ * @param[in] transfer  Pointer to transfer data structure
+ */
+static void stream_rx_xfer_done(usb_transfer_t *in_xfer)
+{
+    assert(in_xfer);
+
+    uac_iface_t *iface = get_interface_by_ep(in_xfer->bEndpointAddress);
```
roma-jam, line 927: Here we take the interface by EP number (Yes, I know that I did that and it is a cp from HID).

Now I know, that it can't be simplified: 
1. Currently, `context = iface->parent` when then there is only the comparison in line 931 and do not use it later in `stream_rx_xfer_done`.
2. We can use this `context` and point it to `iface` directly. And then, during the `xfer_done` callback, we should only verify that `context` is not `NULL`. If it is not `NULL`, we have the interface pointer. 

In that case, it could help us to eliminate `get_interface_by_ep` and simplify `stream_rx_xfer_done` logic.

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -0,0 +1,2141 @@
...
+    switch (in_xfer->status) {
+    case USB_TRANSFER_STATUS_COMPLETED: {
+
+        // if ringbuffer will overflow, notify user to read data
+        size_t data_len = _ring_buffer_get_len(iface->ringbuf);
+        if (data_len + in_xfer->actual_num_bytes >= iface->ringbuf_size) {
+            uac_host_user_interface_callback(iface, UAC_HOST_DEVICE_EVENT_RX_DONE);
+        }
+
+        // if ringbuffer overflow (happens if user not read in above callback), the data will be dropped
```
roma-jam, line 945: How do you think, should we notify user with a callback that some data were dropped?

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -0,0 +1,2141 @@
...
+        uac_host_interface_unlock(uac_iface);
+        return ESP_OK;
+    }
+
+    UAC_ENTER_CRITICAL();
+    if (--uac_iface->parent->opened_cnt == 0) {
+        UAC_EXIT_CRITICAL();
+        UAC_GOTO_ON_ERROR(usb_host_device_close(s_uac_driver->client_handle, uac_iface->parent->dev_hdl), "Unable to close USB device");
+        ESP_LOGD(TAG, "line %d, Close Device addr %d", __LINE__, uac_iface->parent->addr);
+        UAC_GOTO_ON_ERROR(_uac_host_device_delete(uac_iface->parent), "Unable to delete UAC device");
```
roma-jam, line 1780: It seems, that we have: 
1. TAILQ of devices (which is USB device with specific dev_hdl). This is a list of all USB devices, currently connected to the esp. That means, that when device is in this list, it physically attached to the port. 
2. TAILQ of interfaces (which is a uac device, and each of uac has a parent from 1.). 

When use closes the uac device, and it was the only one opened, seems that we don't need to remove the parent. Because physically, the device can be still connected to the port. 
So, it means, that if the we called: `uac_open_dev` - > `uac_close_dev` ->  `uac_open_dev`, w need tp physically reconnect the USB device, because during the third `uac_open_dev` there will be no parent in the list. 

Is that intentional?

roma-jam (commented):
@leeebo Thanks for the such a huge addition. 
In general, the idea is clear. Apparently, I was not able to verify it on a hardware, but probably you will help me and suggest how I can do that. 

I have left couple of thing, that came to me during the review. 

The main idea is: we have three big logical blocks (ringbuffer, USB device handling with its ctrl_transfer handling, interface handling (uac_device part)). How do you think, would it make sense to distinguish them to several separate files?

host/class/uac/usb_host_uac/include/usb/uac_host.h
```diff
@@ -0,0 +1,393 @@
...
+ * @param[in] size            Number of bytes to write
+ * @param[in] timeout         Timeout in ticks. For milliseconds, please use 'pdMS_TO_TICKS()' macros
+ * @return esp_err_t
+ * - ESP_OK on success
+ * - ESP_ERR_INVALID_ARG if the device handle or data is invalid
+ * - ESP_ERR_INVALID_STATE if the device is not in the right state
+ * - ESP_FAIL if write failed or timeout
+*/
+esp_err_t uac_host_device_write(uac_host_device_handle_t uac_dev_handle, uint8_t *data, uint32_t size,
+                                uint32_t timeout);
```
tore-espressif, line 353: Docs: Could you please add a note that would say that these functions return _before_ the USB transfer is complete?

Because, in all other USB Host drivers the read/write operations also take `timeout` parameter, but they return only _after_ the transfer is completed. In UAC we return after that data is added to the ringuf

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -0,0 +1,2141 @@
...
+    }
+    free(uac_iface->iface_alt);
+    free(uac_iface);
+    return ret;
+}
+
+/**
+ * @brief Remove interface from a list
+ *
+ * Use only inside critical section
```
tore-espressif, line 676: Outdated comment? the function enters critical section so this would cause a deadlock

tore-espressif (approved):
LGTM!

host/class/uac/usb_host_uac/uac_host.c
leeebo: @roma-jam make sense, thanks!

leeebo:
> W (19301) uac-host: Feature Unit ID 2 not corresponding to the input terminal
> W (19301) uac-host: Feature Unit ID 7 not corresponding to the input terminal
> assert failed: block_trim_free tlsf.c:496 (block_is_free(block) && "block must be free")

Hi @roma-jam thanks for your test, for the first device, it related to the topic ["should we support a configurable fifo size?"](https://github.com/espressif/esp-usb/pull/27#discussion_r1546252883),

For the second device, it must be something wrong in the descriptor's parsing logic, would you please share the full descriptors?

host/class/uac/usb_host_uac/include/usb/uac_host.h
leeebo: Agree!

host/class/uac/usb_host_uac/uac_host.c
leeebo: From my point of view, it doesn't makes sense to notify data dropped because the audio stream is continuous and the user has no chance to resend the data.

The user should set the appropriate `buffer_threshold` and need to know what `UAC_HOST_DEVICE_EVENT_RX_DONE` means in order to effectively ensure that the data flow is normal.

host/class/uac/usb_host_uac/uac_host.c
leeebo: Hi @roma-jam as the [workflow](https://github.com/espressif/esp-usb/pull/27#issue-2202595629) described.

1. TAILQ of devices, is a list of Opened physical device.
2. TAILQ of interfaces, is a list of Opened logic UAC device

When Open logic UAC device, it will first check if physical device in the list, if not, add the device to the list.

If all logic UAC device closed, the physical device will be removed from list..

leeebo:
> @leeebo sorry I was not aware that the tinyusb supports only UAC2.0.
> 
> In this case, please enable at least build in CI by adding a path to your new component to `.idf_build_apps.toml` file in the root dir.

Hi @peter-marcisovsky would you please point out how to just build the test app without run it?388bad100efac41d80a11236d2769a49c52999b2 seems not work

roma-jam:
Sure, 
> For the second device, it must be something wrong in the descriptor's parsing logic, would you please share the full descriptors?

[jabra_ev_hs.txt](https://github.com/espressif/esp-usb/files/15146592/jabra_ev_hs.txt)

peter-marcisovsky:
Hi @leeebo, to reply to your comment:
> Hi @peter-marcisovsky would you please point out how to just build the test app without run it? [388bad1](https://github.com/espressif/esp-usb/commit/388bad100efac41d80a11236d2769a49c52999b2) seems not work

To build a test app, without running in in the CI, you could mark pytest as [temporarily disabled in CI](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/contribute/esp-idf-tests-with-pytest.html#mark-temporarily-disabled-in-ci)

We are using this approach, for example in usb examples in [esp-idf](https://github.com/espressif/esp-idf/blob/d4cd437ede613fffacc06ac6d6c93a083829022f/examples/peripherals/usb/device/tusb_console/pytest_usb_device_console.py#L14)

host/class/uac/usb_host_uac/include/usb/uac.h
leeebo: Now user can config the value with Kconfig

host/class/uac/usb_host_uac/test_app/pytest_usb_host_uac.py
```diff
@@ -0,0 +1,17 @@
...
+from typing import Tuple
+
+import pytest
+from pytest_embedded_idf.dut import IdfDut
+
+
+@pytest.mark.esp32s2
+@pytest.mark.esp32s3
+@pytest.mark.temp_skip_ci(targets=['esp32s2', 'esp32s3'], reason='unable to mock UAC 1.0 device with tinyusb')
+@pytest.mark.usb_host
```
peter-marcisovsky, line 13: Hopefully the last CI fix:
Please remove the `@pytest.mark.temp_skip_ci` as it only work in the esp-idf CI. 

As an alternative, we can delete the `@pytest.mark.usb_host`, to prevent pytest from running this test case. We can add the proper runner marker, once the CI is extended with the new runners tesing this UAC class.

The pytest called from `build_and_run_test_app_usb.yml` looks only for `usb_host` markers. So it should not collect this test case at all, since it does not have the `usb_host` marker.

```suggestion
# No runner marker, unable to mock UAC 1.0 device with tinyusb
@pytest.mark.esp32s2
@pytest.mark.esp32s3
```

leeebo: Thanks @peter-marcisovsky

leeebo:
> Sure,
> 
> > For the second device, it must be something wrong in the descriptor's parsing logic, would you please share the full descriptors?
> 
> [jabra_ev_hs.txt](https://github.com/espressif/esp-usb/files/15146592/jabra_ev_hs.txt)

Hi @roma-jam, the logic blocks in your UAC device is like the below:

image

However the current [feature unit parsing logic](https://github.com/leeebo/esp-usb/blob/feat/usb_host_uac/host/class/uac/usb_host_uac/uac_host.c#L1188) can not work with `MIXER_UNIT `, I will refactor the parsing logic in this PR to make it more flexible (Replace the one sequential descriptor query with multiple times )

leeebo:
Hi @roma-jam , the latest commit should fixed the `MIXER_UNIT ` handle bug, please try again if you have time.

BTW, would you please try with ESP32-P4 for the `EP MPS (200) exceeds supported limit (128)` issue, it should work.

host/class/uac/usb_host_uac/test_app/sdkconfig.defaults
peter-marcisovsky, line 8: Could you please disable the watchdog for the unity menu?

```suggestion
CONFIG_USB_HOST_HW_BUFFER_BIAS_PERIODIC_OUT=y

# Disable watchdogs, they'd get triggered during unity interactive menu
CONFIG_ESP_INT_WDT=n
CONFIG_ESP_TASK_WDT=n
```

host/class/uac/usb_host_uac/test_app/main/test_host_uac.c
```diff
@@ -0,0 +1,904 @@
...
+            TEST_ASSERT_EQUAL(UAC_DEVICE_EVENT, evt_queue.event_group);
+            uac_host_device_handle_t uac_device_handle = evt_queue.device_evt.handle;
+            uac_host_device_event_t event = evt_queue.device_evt.event;
+            switch (event) {
+            case UAC_HOST_DEVICE_EVENT_TX_DONE:
+                if ((uint32_t)(s_buffer + offset_size) > (uint32_t)wav_file_end) {
+                    s_buffer = (uint16_t *)(wav_file_start + 44);
+                    volume += 20;
+                    TEST_ASSERT_EQUAL(ESP_OK, uac_host_device_set_volume(uac_device_handle, volume > 100 ? 100 : volume));
+                    printf("Volume: %" PRIu32 "\n", volume);
```
peter-marcisovsky, line 555: I tired running the test app, and the volume was set from 10 to 110 on my headphones. 
Is there any reason for that? 

In a comment above (line 541) you are stating that the volume will be set from 10 to 100, which of course will, because of the `uac_device_handle, volume > 100 ? 100 : volume` in the line above. 
But this printf will say the volume is 110.

host/class/uac/usb_host_uac/test_app/main/test_host_uac.c
```diff
@@ -0,0 +1,904 @@
...
+}
+
+TEST_CASE("test uac device handling", "[uac_host][known_device]")
+{
+    test_uac_setup();
+    // handle device connection
+    uint8_t mic_iface_num = 0;
+    uint8_t spk_iface_num = 0;
+    test_handle_dev_connection(&mic_iface_num, NULL);
+    TEST_ASSERT_EQUAL(UAC_DEV_MIC_IFACE_NUM, mic_iface_num);
```
peter-marcisovsky, line 304: I got an error here when running this test: `uac_device_handle, volume > 100 ? 100 : volume`
Do I need any special UAC device for testing? 
I am using Hennheiser pc8 headphones

peter-marcisovsky (commented):
Some observations from running the test app

host/class/uac/usb_host_uac/test_app/main/test_host_uac.c
leeebo: @peter-marcisovsky, you are right. the test case label with `[known_device]` aims to check if interfaces be parsed correctlly.

Sorry for the confusion. I will add a comment there.

host/class/uac/usb_host_uac/test_app/main/test_host_uac.c
leeebo: the printf should also be `printf("Volume: %" PRIu32 "\n", volume > 100 ? 100 : volume);`, fixed.

peter-marcisovsky:
Hi @leeebo 
Just tired the test app again. This time I tried running pytest. There is a following issue: 

```
2024-05-22 11:19:41 |    |   \/   _____/\______   \ _/  |_  ____   _______/  |_ 
2024-05-22 11:19:41 |    |   /\_____  \  |    |  _/ \   __\/ __ \ /  ___/\   __\
2024-05-22 11:19:41 |    |  / /        \ |    |   \  |  | \  ___/ \___ \  |  |  
2024-05-22 11:19:41 |______/ /_______  / |______  /  |__|  \___  >____  > |__|  
2024-05-22 11:19:41                  \/         \/             \/     \/        
2024-05-22 11:19:41 
2024-05-22 11:19:41 
2024-05-22 11:19:41 Press ENTER to see the list of tests.
2024-05-22 11:19:41 [uac_host]
2024-05-22 11:19:41 Running tests matching '[uac_host]'...
2024-05-22 11:19:41 Running test uac device handling...
2024-05-22 11:19:41 I (591) UAC_TEST: USB Host installed
2024-05-22 11:19:41 I (591) UAC_TEST: UAC Class Driver installed
2024-05-22 11:19:42 ./main/test_host_uac.c:308:test uac device handling:FAIL: Expected 3 Was 1. Function [uac_host][known_device]MALLOC_CAP_8BIT: Before 383072 bytes free, After 368080 bytes free (delta -14992)
2024-05-22 11:19:42 
2024-05-22 11:19:42 Running test uac rx reading...
2024-05-22 11:19:42 E (998) usb_phy: usb_new_phy(296): selected PHY is in use
2024-05-22 11:19:42 ./main/test_host_uac.c:151:test uac rx reading:FAIL: Expected 0 Was 259. Function [uac_host][rx]MALLOC_CAP_8BIT: Before 368080 bytes free, After 363252 bytes free (delta -4828)
2024-05-22 11:19:42 
2024-05-22 11:19:42 Running test uac tx writing...
2024-05-22 11:19:42 E (1023) usb_phy: usb_new_phy(296): selected PHY is in use
2024-05-22 11:19:42 ./main/test_host_uac.c:151:test uac tx writing:FAIL: Expected 0 Was 259. Function [uac_host][tx]MALLOC_CAP_8BIT: Before 363252 bytes free, After 358424 bytes free (delta -4828)
```

If a test fails, in my case its the very first test (the issue with known devices which I already pointed out, and you already explained.. so it's not an issue here), all the following tests collected by pytest will fail as well because of
`E (998) usb_phy: usb_new_phy(296): selected PHY is in use`

Could you please release the PHY, once a test fails? 
In the esp-idf usb tests for example, we are calling `usb_del_phy()` during the test teardown.

leeebo:
@peter-marcisovsky 
> Could you please release the PHY, once a test fails? In the esp-idf usb tests for example, we are calling `usb_del_phy()` during the test teardown.

Fixed now, you will get `5 Tests 1 Failures` with your device now.

peter-marcisovsky (approved):
Thanks for the additions. 
LGTM :+1:

leeebo:
Hi @tore-espressif  @roma-jam @peter-marcisovsky  Thanks very much for the review work. I will squash the commits and release the first usb_host_uac v1.0.0.

PR #29, opened 2024-04-16
feat(ci): Build ESP-IDF examples in esp-usb CI

tore-espressif:
Build esp-idf USB device examples in esp-usb CI


roma-jam (approved):
LGTM!

PR #30, opened 2024-04-22
feat(esp_modem_usb_dte): add esp32p4 support & add cat.4 ec20 modem

leeebo:
As esp32p4 has been supported in [usb_host_cdc_acm](https://github.com/espressif/esp-usb/tree/master/host/class/cdc/usb_host_cdc_acm), it should be the right time for us to support higher throughput Cat.4 Cellular module (DL 150 Mbps, UL 50 Mbps).

- [x] Support esp32p4 in esp_modem_usb_dte
- [x] Add Cat.4 module EC20
- [x] Add support in esp_modem example https://github.com/espressif/esp-protocols/pull/556
~~- [ ] Speed iperf test~~

### DIFF
--- a/host/class/cdc/esp_modem_usb_dte/CMakeLists.txt
+++ b/host/class/cdc/esp_modem_usb_dte/CMakeLists.txt
@@ -1,5 +1,4 @@
 idf_component_register(SRCS "esp_modem_usb.cpp" "esp_modem_usb_api_target.cpp" "esp_modem_usb_c_api.cpp"
-                       REQUIRED_IDF_TARGETS esp32s2 esp32s3
                        PRIV_INCLUDE_DIRS "private_include"
                        INCLUDE_DIRS "include")
 
--- a/host/class/cdc/esp_modem_usb_dte/CHANGELOG.md
+++ b/host/class/cdc/esp_modem_usb_dte/CHANGELOG.md
@@ -1,7 +1,11 @@
 ## Unreleased
 
+## 1.2.0
+
 - Fixed C++ build error for `usb_host_config_t` backward compatibility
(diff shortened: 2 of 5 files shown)


host/class/cdc/esp_modem_usb_dte/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 idf_component_register(SRCS "esp_modem_usb.cpp" "esp_modem_usb_api_target.cpp" "esp_modem_usb_c_api.cpp"
-                       REQUIRED_IDF_TARGETS esp32s2 esp32s3
+                       REQUIRED_IDF_TARGETS esp32s2 esp32s3 esp32p4
```
tore-espressif, line 2: This line can be removed. The required target is now taken from https://github.com/espressif/esp-usb/blob/master/host/class/cdc/usb_host_cdc_acm/idf_component.yml#L7

tore-espressif (commented):
@leeebo Thank you for the update!

There is one outstanding issue for some USB modems on P4. Sometimes the esp_modem tries to append received data to USB data buffer, which causes error with unaligned cache sync.

We can merge it now and I can fix it later, what do you think?

david-cermak:
> There is one outstanding issue for some USB modems on P4. Sometimes the esp_modem tries to append received data to USB data buffer, which causes error with unaligned cache sync.

@leeebo @tore-espressif Do you think it's happening only on P4? Probably not the cache sync issue, but maybe those "appended fragments" getting lost?
I hear there're still some stability issues when using USB with `esp_modem` (compared to UART), the latest is here https://github.com/espressif/esp-protocols/issues/589

tore-espressif:
This particular issue is only relevant for P4.

I can check the issue you linked, but no sooner then end of this month

PR #33, opened 2024-04-23
UVC deployment strategy (for discussion)

tore-espressif:
This PR changes our libuvc submodule from upstream https://github.com/libuvc/libuvc to Espressif's fork https://github.com/espressif/libuvc

Please note that this deployment strategy is different from TinyUSB, where we release TinyUSB as one component and TinyUSB additions as a second one. This approach brings a lot of maintenance effort, though. So, IMO, this simplified strategy is more suitable for UVC

Opinions welcome!

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "host/usb_host_uvc/libuvc"]
 	path = host/class/uvc/usb_host_uvc/libuvc
-	url = ../../libuvc/libuvc.git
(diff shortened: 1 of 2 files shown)


leeebo (approved):
LGTM!

roma-jam (approved):
custom fork for libuvc, right?
Any ideas how to maintain the changes in the upstream? We used fixed version though, so maybe it will be no necessary.   
LGTM.

PR #34, opened 2024-05-21
USB Host MSC: Add request sense feature

peter-marcisovsky:
Closing IDF-9890

Added request sense feature for MCS driver.

After each  BOT command execution a `CSW` is checked for possible errors. If an error is present in the MSC device a `bCSWStatus` is returned to the Host as non zero. Ref: [USB Mass Storage Class – Bulk Only Transport](https://www.usb.org/sites/default/files/usbmassbulk_10.pdf) table 5.3. But a current MSC Host driver does not allow the user to find out what is causing the error.

This MR adds a feature to the MSC Host to issue a request sense to the MSC device each time, after checking `CSW` for errors. The request sense command returns specific error code. Ref: [USB Mass Storage Class – UFI Command Specification](https://www.usb.org/sites/default/files/usbmass-ufi10.pdf) Chapter 4.11 and Table 51

### DIFF
--- a/host/class/msc/usb_host_msc/src/msc_scsi_bot.c
+++ b/host/class/msc/usb_host_msc/src/msc_scsi_bot.c
@@ -7,6 +7,7 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include "esp_log.h"
+#include "inttypes.h"
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -67,6 +68,29 @@ static const char *TAG = "USB_MSC_SCSI";
 #define CBW_SIZE        31
 
 #define CWB_FLAG_DIRECTION_IN (1<<7) // device -> host
+
+/**
+ * @brief LUT with error codes and descriptions
+ *
+ * @see USB Mass Storage Class – UFI Command Specification, Revision 1.0
+ * Table 51 - Sense Keys, ASC/ASCQ Listing for All Commands (sorted by Key)
+ *
+ */
+typedef struct {
+    uint8_t sense_key;
+    uint8_t asc;
+    uint8_t ascq;
+    const char *description;
+} sense_errors_t;
+
+const sense_errors_t sense_errors_lut[] = {
+    {0x00, 0x00, 0x00, "NO SENSE"},
+    {0x07, 0x27, 0x00, "WRITE PROTECTED MEDIA"},
+
+    // add more items as needed
+};
+
+#define SENSE_ERROR_COUNT (sizeof(sense_errors_lut) / sizeof(sense_errors_t))
 
 /**
  * @brief Command Block Wrapper structure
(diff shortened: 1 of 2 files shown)


tore-espressif (approved):
LGTM @roma-jam PTAL too

host/class/msc/usb_host_msc/src/msc_scsi_bot.c
```diff
@@ -68,6 +69,21 @@ static const char *TAG = "USB_MSC_SCSI";
...
 
+/**
+ * @brief Sense keys list
+ *
+ * @see USB Mass Storage Class – UFI Command Specification, Revision 1.0
+ * Table 51 - Sense Keys, ASC/ASCQ Listing for All Commands (sorted by Key)
+ *
+ */
+typedef enum {
+    SENSE_WRITE_PROTECTED_MEDIA = 0x00072700,
```
roma-jam, line 80: Nitpick: Maybe it make sense to split it into three enums: Sense key,  ASC and ASCQ or to create a struct of REQUEST_SENSE reply. 
Something like: 

```
struct request_sense_data_t {
    union { 
        struct { 
            uint8_t err_code:   7;
            uint8_t valid:      1;
        }; 
        uint8_t byte0;
    };
    uint8_t byte1_reserved;
    union {
        struct {
            uint8_t sense_key: 4;
            uint8_t reserved4: 4;
        };
        uint8_t byte2;
    };
    uint8_t info[4];
    uint8_t additional_sense_len;
    uint8_t byte8_11_reserved[4];
    uint8_t asc;
    uint8_t ascq;
    uint8_t byte14_17_reserved[4];
} __attribute__((packed));
```
For easier representation of return values and documentation matching
image

host/class/msc/usb_host_msc/src/msc_scsi_bot.c
```diff
@@ -273,6 +291,21 @@ esp_err_t bot_execute_command(msc_device_t *device, msc_cbw_t *cbw, void *data,
     return check_csw(&csw, cbw->tag);
 }
 
+static void decode_sense_keys(cbw_sense_response_t *sense_response)
+{
+    const uint32_t key = 0x00 << 24 |
```
roma-jam, line 296: Then, here you don't need to make all this shifts.

roma-jam (approved):
One nitpick, otherwise LGTM!

host/class/msc/usb_host_msc/src/msc_scsi_bot.c
peter-marcisovsky: Such a struct is already implemented in the `msc_sci_bot.c` as Command Block Wrapper

```
typedef struct __attribute__((packed))
{
    uint8_t error_code;
    uint8_t reserved_0;
    uint8_t sense_key;
    uint32_t info;
    uint8_t sense_len;
    uint32_t reserved_1;
    uint8_t sense_code;
    uint8_t sense_code_qualifier;
    uint32_t reserved_2;
} cbw_sense_response_t;
```

But it's not entirely according to the documentation, unlike what you suggested. I think we can update the CBW structs to be exactly according to the docs, but it would make sense to update all the CBW structs, in different MR.

host/class/msc/usb_host_msc/src/msc_scsi_bot.c
peter-marcisovsky: Thx for the suggestion, I updated the `decode_sense_keys` functions with a lookup table of error codes and their descriptions. PTAL.

host/class/msc/usb_host_msc/src/msc_scsi_bot.c
```diff
@@ -289,7 +329,13 @@ esp_err_t scsi_cmd_read10(msc_host_device_handle_t dev,
         .length = __builtin_bswap16(num_sectors),
     };
 
-    return bot_execute_command(device, &cbw.base, data, num_sectors * sector_size);
+    esp_err_t ret = bot_execute_command(device, &cbw.base, data, num_sectors * sector_size);
+
+    // In case of an error, get an error code
+    if (ret != ESP_OK) {
```
leeebo, line 335: How about `if (unlikely(ret != ESP_OK))`? As the function is called frequently and unlikely to return ESP_OK

peter-marcisovsky: Thx for a suggestion. I have added the `unlikely` macro for issuing of the request sense command.

tore-espressif:
@peter-marcisovsky Please let me know if we should wait for the CI pass or whether I can force merge 🥇

peter-marcisovsky:
@tore-espressif we are only waiting for GitLab master branch sync to GH, to pass the CI. So This MR is ready to be merged

roma-jam:
@peter-marcisovsky @tore-espressif 
CI unlocked, feel free to merge :rocket:

PR #35, opened 2024-05-21
CI: Add build and target run for IDF_v5.3

peter-marcisovsky:
Closes [IDF-9867](https://jira.espressif.com:8443/browse/IDF-9867)

This MR adds IDF 5.3 to the CI build and target run

### DIFF
--- a/.github/workflows/build_and_run_test_app_usb.yml
+++ b/.github/workflows/build_and_run_test_app_usb.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


tore-espressif (approved):
Thanks!

roma-jam (approved):
During adding the CI examples build verification, @Dazza0 proposed very nice idea to think about trying to use wildcards in such case (https://github.com/espressif/tinyusb/pull/24#discussion_r1576855388)

It will help us to exclude the CI changes during upcoming releases of esp-idf.

Otherwise. LGTM !

peter-marcisovsky:
@roma-jam thx for pointing out. Will take a look if we can implement it.

PR #36, opened 2024-05-21
ci: remove libtinfo5 installation

peter-marcisovsky:
Following up with a fix from [idf-extra-components](https://github.com/espressif/idf-extra-components/pull/333)


igrr:
@peter-marcisovsky FYI, I've reverted the update of clang-tidy-sarif. There is an issue with duplicate entries in the generated SARIF file. I'll submit a fix to clang-tidy-sarif and upgrade it later.

igrr (approved):

roma-jam (approved):
LGTM!

PR #39, opened 2024-06-11
feature(esp_tinyusb): Added tusb_teardown() call while tinyusb_driver_uninstall() 

roma-jam:
## Requirements
- To uninstall tinyusb driver, a teardown tinyusb stack should be called. 
- `tud_deinit()` feature was implemented in upstream, starting with version v0.17.0.

## Description
- Added `tusb_teardown()` call while `tinyusb_driver_uninstall()` 
- Added test app to verify teardown feature

## Testing
 1. Install TinyUSB device without any class
 2. Wait SetConfiguration() (tud_mount_cb)
 3. If attempts == 0 goto step 8
 4.   Wait TEARDOWN_DEVICE_DETACH_DELAY_MS
 5.   Uninstall TinyUSB device
 6.   Wait TEARDOWN_DEVICE_INIT_DELAY_MS
 7.   Decrease attempts by 1, goto step 3
 8. Wait TEARDOWN_DEVICE_DETACH_DELAY_MS
 9. Uninstall TinyUSB device

Notes: 
- when amount of teardown configured to 0, then only one install-uninstall routine is fulfilled. 
- device presence used subprocess call "lsusb" and check the Espressif VID/PIN in the output

## Related issues
- Original `espressif/tinyusb` PR: https://github.com/espressif/tinyusb/pull/27 (closed, because of synchronizing with upstream v0.17)
- Requires tusb_teardown() API: https://github.com/espressif/tinyusb/pull/44
- Resolve https://github.com/espressif/esp-idf/issues/13788
- Closes https://github.com/espressif/esp-idf/issues/15122

### DIFF
--- a/device/esp_tinyusb/test_apps/teardown_device/main/test_teardown.c
+++ b/device/esp_tinyusb/test_apps/teardown_device/main/test_teardown.c
@@ -1,0 +1,142 @@
+/*
+ * SPDX-FileCopyrightText: 2024 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "soc/soc_caps.h"
+#if SOC_USB_OTG_SUPPORTED
+
+//
+#include <stdio.h>
+#include <string.h>
+//
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
+#include "freertos/semphr.h"
+//
+#include "esp_system.h"
+#include "esp_log.h"
+#include "esp_err.h"
+//
+#include "unity.h"
+#include "tinyusb.h"
+#include "tusb_cdc_acm.h"
+
+static const char *TAG = "teardown";
+
+SemaphoreHandle_t wait_mount = NULL;
+
+#define TEARDOWN_DEVICE_INIT_DELAY_MS       1000
+#define TEARDOWN_DEVICE_ATTACH_TIMEOUT_MS   1000
+#define TEARDOWN_DEVICE_DETACH_DELAY_MS     1000
+
+#define TEARDOWN_AMOUNT                     10
+
+#define TUSB_DESC_TOTAL_LEN                 (TUD_CONFIG_DESC_LEN)
+static uint8_t const test_configuration_descriptor[] = {
+    // Config number, interface count, string index, total length, attribute, power in mA
+    TUD_CONFIG_DESCRIPTOR(1, 0, 0, TUSB_DESC_TOTAL_LEN, TUSB_DESC_CONFIG_ATT_SELF_POWERED | TUSB_DESC_CONFIG_ATT_REMOTE_WAKEUP, 100),
+};
+
+static const tusb_desc_device_t test_device_descriptor = {
+    .bLength = sizeof(test_device_descriptor),
+    .bDescriptorType = TUSB_DESC_DEVICE,
+    .bcdUSB = 0x0200,
+    .bDeviceClass = TUSB_CLASS_MISC,
+    .bDeviceSubClass = MISC_SUBCLASS_COMMON,
+    .bDeviceProtocol = MISC_PROTOCOL_IAD,
+    .bMaxPacketSize0 = CFG_TUD_ENDPOINT0_SIZE,
+    .idVendor = 0x303A, // This is Espressif VID. This needs to be changed according to Users / Customers
+    .idProduct = 0x4002,
+    .bcdDevice = 0x100,
+    .iManufacturer = 0x01,
+    .iProduct = 0x02,
+    .iSerialNumber = 0x03,
+    .bNumConfigurations = 0x01
+};
+
+#if (TUD_OPT_HIGH_SPEED)
+static const tusb_desc_device_qualifier_t device_qualifier = {
+    .bLength = sizeof(tusb_desc_device_qualifier_t),
+    .bDescriptorType = TUSB_DESC_DEVICE_QUALIFIER,
+    .bcdUSB = 0x0200,
+    .bDeviceClass = TUSB_CLASS_MISC,
+    .bDeviceSubClass = MISC_SUBCLASS_COMMON,
+    .bDeviceProtocol = MISC_PROTOCOL_IAD,
+    .bMaxPacketSize0 = CFG_TUD_ENDPOINT0_SIZE,
+    .bNumConfigurations = 0x01,
+    .bReserved = 0
+};
+#endif // TUD_OPT_HIGH_SPEED
+
+// Invoked when device is mounted
+void tud_mount_cb(void)
+{
+    xSemaphoreGive(wait_mount);
+}
+
+/**
+ * @brief TinyUSB Teardown specific testcase
+ *
+ * Scenario:
+ * 1. Install TinyUSB device without any class
+ * 2. Wait SetConfiguration() (tud_mount_cb)
+ * 3. If attempts == 0 goto step 8
+ * 4.   Wait TEARDOWN_DEVICE_DETACH_DELAY_MS
+ * 5.   Uninstall TinyUSB device
+ * 6.   Wait TEARDOWN_DEVICE_INIT_DELAY_MS
+ * 7.   Decrease attempts by 1, goto step 3
+ * 8. Wait TEARDOWN_DEVICE_DETACH_DELAY_MS
+ * 9. Uninstall TinyUSB device
+ */
+TEST_CASE("tinyusb_teardown", "[esp_tinyusb][teardown]")
+{
+    wait_mount = xSemaphoreCreateBinary();
+    TEST_ASSERT_NOT_EQUAL(NULL, wait_mount);
+
+    // TinyUSB driver configuration
+    const tinyusb_config_t tusb_cfg = {
+        .device_descriptor = &test_device_descriptor,
+        .string_descriptor = NULL,
+        .string_descriptor_count = 0,
+        .external_phy = false,
+#if (TUD_OPT_HIGH_SPEED)
+        .fs_configuration_descriptor = test_configuration_descriptor,
+        .hs_configuration_descriptor = test_configuration_descriptor,
+        .qualifier_descriptor = &device_qualifier,
+#else
+        .configuration_descriptor = test_configuration_descriptor,
+#endif // TUD_OPT_HIGH_SPEED
+    };
+
+    TEST_ASSERT_EQUAL(ESP_OK, tinyusb_driver_install(&tusb_cfg));
+    // Wait for the usb event
+    ESP_LOGD(TAG, "wait mount...");
+    TEST_ASSERT_EQUAL(pdTRUE, xSemaphoreTake(wait_mount, pdMS_TO_TICKS(TEARDOWN_DEVICE_ATTACH_TIMEOUT_MS)));
+    ESP_LOGD(TAG, "mounted");
+
+    // Teardown routine
+    int attempts = TEARDOWN_AMOUNT;
+    while (attempts--) {
+        // Keep device attached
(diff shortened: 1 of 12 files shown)


device/esp_tinyusb/tinyusb.c
```diff
@@ -69,5 +69,6 @@ esp_err_t tinyusb_driver_install(const tinyusb_config_t *config)
 esp_err_t tinyusb_driver_uninstall()
 {
     tinyusb_free_descriptors();
+    tusb_teardown();
```
roma-jam, line 72: @tore-espressif ,

*when you will have time*

Is there any good way how I can make this change backward compatible with previous `espressif/tinyusb` components (previous versions, which doesn't have this implementation)? 
Will appreciate any suggestions.

device/esp_tinyusb/test_app/main/test_descriptors_config.c
```diff
@@ -232,4 +232,262 @@ TEST_CASE("descriptors_config_all_configured", "[esp_tinyusb][usb_device]")
...
+        .string_descriptor_count = sizeof(test_msc_string_descriptor) / sizeof(test_msc_string_descriptor[0]),
+        .configuration_descriptor = test_msc_fs_configuration_descriptor,
+#if (TUD_OPT_HIGH_SPEED)
+        .hs_configuration_descriptor = test_msc_hs_configuration_descriptor,
+#endif // TUD_OPT_HIGH_SPEED
+    },
+};
+
+
+TEST_CASE("device_class_reconfiguration", "[usb_device]")
```
roma-jam, line 463: @peter-marcisovsky 

After adding the feature from here: https://github.com/espressif/tinyusb/pull/27, we will be able to run `esp_tinyusb` device tests, because they could be run one by one in a one test launch. 

Currently, I have no idea how to add them to our CI and where, but if you have some ideas, please feel free to share. 
Thanks.

roma-jam:
@tore-espressif , @peter-marcisovsky 
This changes unlock the possibility to run usb device tests one by one (https://github.com/espressif/tinyusb/pull/27 should be merged first, though).

I posted two questions for both of you, because it seems that we can make it better and we can use it.

I need that for enumeration driver testing (but maybe I will eliminate esp_tinyusb component from the chain, but anyway), so feel free to check it out when possible. 

Meanwhile, I will go and cover the Enum Drivier as much as I can.

device/esp_tinyusb/tinyusb.c
roma-jam: Here are some ideas, that came to me: 
1. Tie it to the configuration and use idf version to determine the necessity and implementation of tusb_teardown()
2. Define it somewhere in esp_tinyusb with WEAK parameter. Which let us redefine in with new version and don't panic with old versions, when there is no implementation of it
3. Enable it with menuconfig (which seems really strange for me, but anyway maybe it is possible).

I like option No.2, but maybe there are other simple and better solution.

device/esp_tinyusb/test_app/main/test_descriptors_config.c
peter-marcisovsky: To add usb device test, you just need to create a new pytest marker in your pytest*.py file, like so:

`@pytest.mark.usb_device`

Then, run pytest from `build_and_run_test_app_usb.yml` file with this new marker. 
Right now, we are using only `-m usb_host` marker. Thus there will be one more marker `-m usb_device`
So the new pytest command would look like:
`pytest --target=${{ matrix.idf_target }} -m usb_host -m usb_device --build-dir=build_${{ matrix.idf_target }}`

This should be all.

device/esp_tinyusb/tinyusb.c
```diff
@@ -66,8 +78,18 @@ esp_err_t tinyusb_driver_install(const tinyusb_config_t *config)
     return ESP_OK;
 }
 
-esp_err_t tinyusb_driver_uninstall()
+esp_err_t tinyusb_driver_uninstall(void)
```
peter-marcisovsky, line 81: Could we call this function from `tearDown()`?
Once a test case fails, the PHY does not get released, thus all the following test cases would fail because of the:

```
E (23549) usb_phy: usb_new_phy(296): selected PHY is in use
E (23559) TinyUSB: tinyusb_driver_install(65): Install USB PHY failed
```

peter-marcisovsky (commented):
Left some observations from using the test app

device/esp_tinyusb/tinyusb.c
```diff
@@ -66,8 +78,18 @@ esp_err_t tinyusb_driver_install(const tinyusb_config_t *config)
...
+    esp_err_t ret = tusb_stop_task();
+
+    if (ret != ESP_OK) {
+        return ret;
+    }
+
+    if (!tusb_teardown()) {
+        return ESP_ERR_NOT_FINISHED;
+    }
+
```
leeebo, line 92: @roma-jam Should we restore the USB to the default function (USB-Serial-JTAG) at the end of uninstall?

roma-jam: @leeebo I don't think that would be a good idea, as for user it is transparent in that case. 
Even, IDK, whether we have such feature before installing tinyUSB driver? 

Just in case: do we have such call (maybe in rom) to enable this feature? Or you thinking about configure the PHY during teardown in the way, to enable USB-Serial-JTAG?

leeebo: The USB-Serial-JTAG is a hardware circuit that can respond to the host's requests and is enabled by default. Users can use the USB-Serial-JTAG for USB download without any code in APP.

But after tinyUSB install then uninstalled, the USB phy not be switched back to USB-Serial-JTAG

greenaddress: @roma-jam @leeebo as a user of the library in esp32s3 I think teardown/uninstall should not auto-reinstall jtag but there should be a convenience function that allows to restore it, if necessary it could be called automatically and be driven by Kconfig to provide best defaults. In my use case i want to be able to switch from usb cdc guest to msc usb host and viceversa and that is currently not possible without a device restart

finger563:
Is there a timeline for this feature to be finalized?

roma-jam:
Hi @finger563 ,

This feature allows the re-configure the driver with these changes: https://github.com/espressif/tinyusb/pull/27
But as far as the same changes are already in the upstream, so we will rebase these changes after updating version to v0.17.0.
For current feature there is no available timeline for now, but the one thing for sure: the update to v0.17.0 will be done first.

Meanwhile, this feature should be available in the upstream of TinyUSB, if anything it is possible to use it. 

Sorry for the inconvenience.

roma-jam:
@tore-espressif @peter-marcisovsky 
This is the enabling of teardown feature via `tinyusb_driver_uninstall()` call from esp_tinyusb wrapper. 

To enable the CI (I am expecting the teardown_device test to fail), it is required to release `espressif/tinyusb` component with changes from here: https://github.com/espressif/tinyusb/pull/44

Feel free to review the changes in current PR and I will relaunch CI after the tinyusb release (Which I planned on 22th of December, but will do that next Monday, 16th).

device/esp_tinyusb/test_apps/teardown_device/pytest_teardown_device.py
```diff
@@ -0,0 +1,92 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+from pytest_embedded_idf.dut import IdfDut
+import serial
+# import serial.tools.list_ports
+from serial.tools.list_ports import comports
+# import time
+from time import sleep
+from time import time
```
peter-marcisovsky, line 11:
```suggestion
from time import sleep, time
```

peter-marcisovsky (commented):
Leaving some comments. 
I am having some issues with running the test locally. Will try to fix it an maybe add something else to the review.

device/esp_tinyusb/test_apps/teardown_device/pytest_teardown_device.py
```diff
@@ -0,0 +1,92 @@
...
+
+    for i in range(ITERATIONS):
+        print(f"Iteration {i+1} of {ITERATIONS}")
+
+        # Wait for the device to appear
+        print("Waiting for the device to connect...")
+        tusb_cdc = wait_for_tusb_cdc(TUSB_VID, TUSB_PID)
+        if not tusb_cdc:
+            print("Error: Device did not appear within the timeout period.")
+            break
```
peter-marcisovsky, line 54: Maybe, consider adding assert here as well. Otherwise, if the device does not appear, the test passes, which is incorrect.

```suggestion
        if not tusb_cdc:
            print("Error: Device did not appear within the timeout period.")
            assert True
```

device/esp_tinyusb/test_apps/teardown_device/main/test_teardown.c
```diff
@@ -0,0 +1,167 @@
...
+    .bcdDevice = 0x0100,
+    .iManufacturer = 0x01,
+    .iProduct = 0x02,
+    .iSerialNumber = 0x03,
+    .bNumConfigurations = 0x01
+};
+
+static const uint16_t cdc_desc_config_len = TUD_CONFIG_DESC_LEN + CFG_TUD_CDC * TUD_CDC_DESC_LEN;
+static const uint8_t cdc_desc_configuration[] = {
+    TUD_CONFIG_DESCRIPTOR(1, 4, 0, cdc_desc_config_len, TUSB_DESC_CONFIG_ATT_REMOTE_WAKEUP, 100),
```
peter-marcisovsky, line 49: We have only 2 interfaces, not 4, correct?

tore-espressif: It should be `CFG_TUD_CDC * 2` (2 interfaces per CDC class). Equals 2 in this configuration

device/esp_tinyusb/test_apps/teardown_device/main/test_teardown.c
```diff
@@ -0,0 +1,167 @@
...
+
+#include "soc/soc_caps.h"
+// #if SOC_USB_OTG_SUPPORTED
+
+//
+#include <stdio.h>
+#include <string.h>
+//
+#include "freertos/FreeRTOS.h"
+#include "freertos/task.h"
```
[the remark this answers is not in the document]
roma-jam: Hm...do I really need one.. :thinking:

device/esp_tinyusb/test_apps/teardown_device/pytest_teardown_device.py
```diff
@@ -0,0 +1,92 @@
...
+
+        try:
+            # Open the serial port
+            with serial.Serial(port=tusb_cdc, baudrate=9600, timeout=1) as cdc:
```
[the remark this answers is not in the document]
roma-jam: Yeah, that is right. It also could be simplified to the line: 
`with Serial(s[0]) as cdc0:`

I'll think about simplifying it and use the common approach for all tests.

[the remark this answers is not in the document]
roma-jam: I curved out the cdc teardown test to another PR: https://github.com/espressif/esp-usb/pull/106
And there is no pyserial in the pytest. I changed it to pyusb. 

So the `serial.Serial` is not relevant for this PR anymore.

roma-jam:
I have updated the test, changing the logic of checking the result of "lsusb" output. 
The previous test with teardown + cdc data exchange between host and device moved here: https://github.com/espressif/esp-usb/pull/106

PTAL

tore-espressif (approved):
@roma-jam Looks good, thank you for the test!

roma-jam:
Applied several fixes from review in here: https://github.com/espressif/esp-usb/pull/106

peter-marcisovsky (approved):
LGTM, I could run the test locally on my setup, works fine.

PR #40, opened 2024-06-12
fix issue that out_xfer unintentionally returned by previous transmission (IEC-121)

radiotommy:
for timeout or any reason the semaphore could left untaken before the new transmission start. it causing the transmission pick up the previous actual_num_bytes and return failure because of incorrect number of bytes transferred.

### DIFF
--- a/host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
+++ b/host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
@@ -1155,15 +1155,27 @@ esp_err_t cdc_acm_host_data_tx_blocking(cdc_acm_dev_hdl_t cdc_hdl, const uint8_t
     memcpy(cdc_dev->data.out_xfer->data_buffer, data, data_len);
     cdc_dev->data.out_xfer->num_bytes = data_len;
     cdc_dev->data.out_xfer->timeout_ms = timeout_ms;
+    cdc_dev->data.out_xfer->actual_num_bytes = 0;
     ESP_GOTO_ON_ERROR(usb_host_transfer_submit(cdc_dev->data.out_xfer), unblock, TAG,);
 
     // Wait for OUT transfer completion
-    taken = xSemaphoreTake((SemaphoreHandle_t)cdc_dev->data.out_xfer->context, pdMS_TO_TICKS(timeout_ms));
-    if (!taken) {
-        // Reset the endpoint
-        cdc_acm_reset_transfer_endpoint(cdc_dev->dev_hdl, cdc_dev->data.out_xfer);
-        ret = ESP_ERR_TIMEOUT;
-        goto unblock;
+
+    uint32_t start = xTaskGetTickCount();
+    uint32_t timeout_ticks = pdMS_TO_TICKS(timeout_ms);
+
+    while(cdc_dev->data.out_xfer->actual_num_bytes < data_len) {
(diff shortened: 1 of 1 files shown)


tore-espressif:
Thank you @radiotommy for the PR, I'll have a look ASAP

radiotommy:
@tore-espressif 
The test run failure seems not caused by this change, is anyone looking into it?

tore-espressif:
> The test run failure seems not caused by this change, is anyone looking into it?

Yes, we got a regression few days ago that should not be caused by your changes. 

We had a slightly different proposal to fix this issue. Let me check with my colleagues and I'll get back to you

host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
```diff
@@ -1155,15 +1155,27 @@ esp_err_t cdc_acm_host_data_tx_blocking(cdc_acm_dev_hdl_t cdc_hdl, const uint8_t
...
-    if (!taken) {
-        // Reset the endpoint
-        cdc_acm_reset_transfer_endpoint(cdc_dev->dev_hdl, cdc_dev->data.out_xfer);
-        ret = ESP_ERR_TIMEOUT;
-        goto unblock;
+
+    uint32_t start = xTaskGetTickCount();
+    uint32_t timeout_ticks = pdMS_TO_TICKS(timeout_ms);
+
+    while (cdc_dev->data.out_xfer->actual_num_bytes < data_len) {
```
tore-espressif, line 1166: Could you please explain the reason for this loop?

I can see two cases:
1. The transfer finishes with `actual_num_bytes == data_len` -> all OK, transfer finished.
2. The transfer finishes with `actual_num_bytes < data_len` -> not OK. In this case we try to take the semaphore again, but we will never get it because we did not resubmit the transfer. So we end up with `ESP_ERR_TIMEOUT`, same as before.

Could you please describe the bug (or provide a link to a GH issue) that you are trying to fix? It could help me understand better.

tore-espressif: > for timeout or any reason the semaphore could left untaken before the new transmission start

To workaround this issue, we could 
``` c
xSemaphoreTake((SemaphoreHandle_t)cdc_dev->data.out_xfer->context, 0);
```
before we submit the transfer. If we set timeout to 0, the function will immediately return and we will make sure that the semaphore is not given from previous transfer. Does that make sense?

tore-espressif: Even better solution is to 
``` c
if (!taken) {
        // Reset the endpoint
        cdc_acm_reset_transfer_endpoint(cdc_dev->dev_hdl, cdc_dev->data.out_xfer);
        xSemaphoreTake((SemaphoreHandle_t)cdc_dev->data.out_xfer->context,  portMAX_DELAY);
        ret = ESP_ERR_TIMEOUT;
        goto unblock;
```

Here we take the semaphore only if the transfer really timed-out

tore-espressif:
Related to https://github.com/espressif/esp-protocols/issues/514

host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
radiotommy: Great idea. I will give it a test on my board this weekend.

PR #41, opened 2024-06-14
fix(host/cdc): Fix TX timeout reaction

tore-espressif:
After TX transfer timeout we reset the endpoint which causes all in-flight transfers to complete. The transfer finished callback gives a transfer finished semaphore which we must take, so we would not affect next TX transfers.

Test included

Closes https://github.com/espressif/esp-protocols/issues/514
Closes https://github.com/espressif/esp-usb/pull/40
Closes https://github.com/espressif/esp-idf/issues/13797

### DIFF
--- a/host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
+++ b/host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
@@ -574,7 +574,6 @@ static esp_err_t cdc_acm_transfers_allocate(cdc_dev_t *cdc_dev, const usb_ep_des
     cdc_dev->ctrl_transfer->timeout_ms = 1000;
     cdc_dev->ctrl_transfer->bEndpointAddress = 0;
     cdc_dev->ctrl_transfer->device_handle = cdc_dev->dev_hdl;
-    cdc_dev->ctrl_transfer->context = cdc_dev;
     cdc_dev->ctrl_transfer->callback = out_xfer_cb;
     cdc_dev->ctrl_transfer->context = xSemaphoreCreateBinary();
     ESP_GOTO_ON_FALSE(cdc_dev->ctrl_transfer->context, ESP_ERR_NO_MEM, err, TAG,);
@@ -1152,16 +1151,19 @@ esp_err_t cdc_acm_host_data_tx_blocking(cdc_acm_dev_hdl_t cdc_hdl, const uint8_t
     }
 
     ESP_LOGD("CDC_ACM", "Submitting BULK OUT transfer");
+    SemaphoreHandle_t transfer_finished_semaphore = (SemaphoreHandle_t)cdc_dev->data.out_xfer->context;
+    xSemaphoreTake(transfer_finished_semaphore, 0); // Make sure the semaphore is taken before we submit new transfer
+
     memcpy(cdc_dev->data.out_xfer->data_buffer, data, data_len);
     cdc_dev->data.out_xfer->num_bytes = data_len;
     cdc_dev->data.out_xfer->timeout_ms = timeout_ms;
     ESP_GOTO_ON_ERROR(usb_host_transfer_submit(cdc_dev->data.out_xfer), unblock, TAG,);
 
     // Wait for OUT transfer completion
-    taken = xSemaphoreTake((SemaphoreHandle_t)cdc_dev->data.out_xfer->context, pdMS_TO_TICKS(timeout_ms));
+    taken = xSemaphoreTake(transfer_finished_semaphore, pdMS_TO_TICKS(timeout_ms));
     if (!taken) {
-        // Reset the endpoint
-        cdc_acm_reset_transfer_endpoint(cdc_dev->dev_hdl, cdc_dev->data.out_xfer);
+        cdc_acm_reset_transfer_endpoint(cdc_dev->dev_hdl, cdc_dev->data.out_xfer); // Resetting the endpoint will cause all in-progress transfers to complete
+        ESP_LOGW(TAG, "TX transfer timeout");
         ret = ESP_ERR_TIMEOUT;
         goto unblock;
     }
--- a/host/class/cdc/usb_host_cdc_acm/test_app/main/test_cdc_acm_host.c
+++ b/host/class/cdc/usb_host_cdc_acm/test_app/main/test_cdc_acm_host.c
@@ -573,7 +573,6 @@ TEST_CASE("functional_descriptor", "[cdc_acm]")
  */
 TEST_CASE("closing", "[cdc_acm]")
 {
-    nb_of_responses = 0;
     cdc_acm_dev_hdl_t cdc_dev = NULL;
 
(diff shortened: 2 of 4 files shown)


tore-espressif:
@radiotommy I'd like to close your #40 in favor of this PR. Could you please check that this fix works for you?
@wishinlife This should fix your https://github.com/espressif/esp-protocols/issues/514

@david-cermak PTAL too, we'd like to release `usb_host_cdc_acm` v2.0.3 ASAP. Are there any changes required to fix https://github.com/espressif/esp-protocols/issues/589 ? Or can everyhing be implemented in `esp_modem_usb_dte`? (these are 2 separate components)

tore-espressif:
@peter-marcisovsky we discussed similar issue earlier today :)

radiotommy:
just tested it on my board. It fixes my problem as well. 
@tore-espressif, no problem, I like yours better :). hope this get merged ASAP

tore-espressif:
> I assume that it's been a design decision to return an error when trying to transmit more bytes than the configured Tx buffer rather then iterate over the bigger size until we transmit everything

@david-cermak yes, this is intentional. We will consider this feature in the future ;)

host/class/cdc/usb_host_cdc_acm/test_app/main/test_cdc_acm_host.c
```diff
@@ -599,6 +599,40 @@ TEST_CASE("closing", "[cdc_acm]")
     vTaskDelay(20); //Short delay to allow task to be cleaned up
 }
 
+/* Basic test to check CDC driver reaction to TX timeout */
+TEST_CASE("tx_timeout", "[cdc_acm]")
+{
+    nb_of_responses = 0;
```
peter-marcisovsky, line 605: We are not checking this value anywhere in this test case. Can we remove it, not to confuse anyone?
Same comment from the test case above, probably a copy-paste of the test case.

host/class/cdc/usb_host_cdc_acm/test_app/main/test_cdc_acm_host.c
```diff
@@ -599,6 +599,40 @@ TEST_CASE("closing", "[cdc_acm]")
...
+        .data_cb = handle_rx,
+        .user_arg = tx_buf,
+    };
+
+    printf("Opening CDC-ACM device\n");
+    TEST_ASSERT_EQUAL(ESP_OK, cdc_acm_host_open(0x303A, 0x4002, 0, &dev_config, &cdc_dev)); // 0x303A:0x4002 (TinyUSB Dual CDC device)
+    TEST_ASSERT_NOT_NULL(cdc_dev);
+    vTaskDelay(10);
+
+    // TX some date with timeout_ms=0. This will cause a timeout
```
peter-marcisovsky, line 623:
```suggestion
    // TX some data with timeout_ms=0. This will cause a timeout
```

host/class/cdc/usb_host_cdc_acm/test_app/main/test_cdc_acm_host.c
```diff
@@ -599,6 +599,40 @@ TEST_CASE("closing", "[cdc_acm]")
...
+    TEST_ASSERT_EQUAL(ESP_ERR_TIMEOUT, cdc_acm_host_data_tx_blocking(cdc_dev, tx_buf, sizeof(tx_buf), 0));
+    vTaskDelay(100); // Wait before trying new TX
+
+    TEST_ASSERT_EQUAL(ESP_OK, cdc_acm_host_data_tx_blocking(cdc_dev, tx_buf, sizeof(tx_buf), 1000));
```
peter-marcisovsky, line 627:
```suggestion
    // TX some data again with higher timeout. This will check a normal operation
    TEST_ASSERT_EQUAL(ESP_OK, cdc_acm_host_data_tx_blocking(cdc_dev, tx_buf, sizeof(tx_buf), 1000));
```
Just to add some comment, that now we want the transfer to pass

host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
```diff
@@ -1158,10 +1157,12 @@ esp_err_t cdc_acm_host_data_tx_blocking(cdc_acm_dev_hdl_t cdc_hdl, const uint8_t
...
 
     // Wait for OUT transfer completion
-    taken = xSemaphoreTake((SemaphoreHandle_t)cdc_dev->data.out_xfer->context, pdMS_TO_TICKS(timeout_ms));
+    SemaphoreHandle_t transfer_finished_semaphore = (SemaphoreHandle_t)cdc_dev->data.out_xfer->context;
+    taken = xSemaphoreTake(transfer_finished_semaphore, pdMS_TO_TICKS(timeout_ms));
     if (!taken) {
-        // Reset the endpoint
-        cdc_acm_reset_transfer_endpoint(cdc_dev->dev_hdl, cdc_dev->data.out_xfer);
+        cdc_acm_reset_transfer_endpoint(cdc_dev->dev_hdl, cdc_dev->data.out_xfer); // Resetting the endpoint will cause all in-progress transfers to complete
+        xSemaphoreTake(transfer_finished_semaphore, pdMS_TO_TICKS(1000)); // The transfer callback gives the semaphore, take it back
```
peter-marcisovsky, line 1164:
```suggestion
        xSemaphoreTake(transfer_finished_semaphore, pdMS_TO_TICKS(timeout_ms)); // The transfer callback gives the semaphore, take it back
```
Do we want to wait here for `timeout_ms` amount, or some fixed value?

peter-marcisovsky (approved):
Just a few minor comments. LGTM.

host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
tore-espressif: The `timeout_ms` applies to regular transfer timeout. 

This is timeout of channel halt. This should happen 'immediately'. But because the user might run this from a task that has higher priority than the CDC driver task, we wait until we get the 'channel halted' interrupt.

So I think it should be OK like this

tore-espressif:
@peter-marcisovsky thanks for the review!

roma-jam (approved):
Just one nitpick about the constant value in the code. 
Otherwise, LGTM.

host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
radiotommy: I think taking the semaphore with 0 timeout just before submit the xfer can avoid this kind of confusion,  also more bug proof.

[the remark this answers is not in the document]
tore-espressif: Yep, it should be triggered 'immediately' in case the USB client task has higher priority than current task.

OK, will fix

tore-espressif: OK, changed!

PR #42, opened 2024-06-17
feat(esp_modem_usb_dte): add esp32p4 support & add cat.4 ec20 modem

tore-espressif:
Closes https://github.com/espressif/esp-usb/pull/30

### DIFF
--- a/host/class/cdc/esp_modem_usb_dte/CHANGELOG.md
+++ b/host/class/cdc/esp_modem_usb_dte/CHANGELOG.md
@@ -1,11 +1,12 @@
-## Unreleased
-
 ## 1.2.0
 
 - Fixed C++ build error for `usb_host_config_t` backward compatibility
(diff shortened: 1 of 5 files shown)


tore-espressif:
@leeebo I cherry picked your commit from #30 and added a 'known issue' section. We can merge ASAP so we don't block any esp_modem examples updates

leeebo:
@tore-espressif Sorry for leaving PR #30 so long, because when I tried to do iperf test (PC->ETH->USB->4G), I encountered some problems (Ping was normal, but TCP communication could not be established when I visited the web page), and I haven't confirmed the root cause of the problem ( NAPT? Or maybe because of P4 Cache?) But most likely 'esp_modem_usb_dte' is unrelated, so I think yes this PR can be merged first

leeebo (approved):
LGTM!

PR #47, opened 2024-07-24
feat(cdc): fix compilation with GCC14 (IEC-138)

Lapshin:
After upgrading to gcc14 i faced this issue: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116070

There is a solution to suppress the errors, but I don't know if you happy with these changes and why the default gnu++XX standard derived from idf could not be used...

P.s.: This change looks valid to me if take into account: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116071#c2

### DIFF
--- a/host/class/cdc/usb_host_vcp/CMakeLists.txt
+++ b/host/class/cdc/usb_host_vcp/CMakeLists.txt
@@ -5,4 +5,7 @@ set_target_properties(${COMPONENT_LIB} PROPERTIES
     CXX_STANDARD 14
     CXX_STANDARD_REQUIRED ON
 )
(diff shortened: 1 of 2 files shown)


igrr:
Note that you probably also need to bump the component version (https://github.com/espressif/esp-usb/blob/5c9b666df35ec4ee7d345123987b64bfa420c7e8/host/class/cdc/usb_host_vcp/idf_component.yml#L2) so that the changes are released.

tore-espressif (approved):

tore-espressif:
Thanks for the update @Lapshin ! I'll merge as is now. 
There is a small refactor of the CDC in progress, I'll revisit this topic during the refactor

PR #57, opened 2024-08-13
fix(host/cdc_acm): Fix CTRL transfer size for devices with small MaxPacketSize0

tore-espressif:
Some devices have MaxPacketSize0 = 8 which is not enough for some CDC specific CTRL requests

Closes https://github.com/espressif/esp-idf/issues/14345

### DIFF
--- a/host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
+++ b/host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
@@ -24,10 +24,14 @@ static const char *TAG = "cdc_acm";
 // bDeviceClass = 0xEF (Miscellaneous Device Class), bDeviceSubClass = 0x02 (Common Class), bDeviceProtocol = 0x01 (Interface Association Descriptor),
 // or when bDeviceClass, bDeviceSubClass, and bDeviceProtocol are 0x00 (Null class code triple), as per https://www.usb.org/defined-class-codes, "Base Class 00h (Device)" section
 // @see USB Interface Association Descriptor: Device Class Code and Use Model rev 1.0, Table 1-1
-#define USB_SUBCLASS_NULL        0x00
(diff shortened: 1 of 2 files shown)


tore-espressif:
@roma-jam @peter-marcisovsky PTAL when you're free :)

host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
```diff
@@ -24,11 +24,15 @@ static const char *TAG = "cdc_acm";
...
-#define USB_SUBCLASS_NULL        0x00
+#define USB_SUBCLASS_NULL          0x00
 #define USB_SUBCLASS_COMMON        0x02
-#define USB_PROTOCOL_NULL    0x00
+#define USB_PROTOCOL_NULL          0x00
 #define USB_DEVICE_PROTOCOL_IAD    0x01
 
+// Control transfer constants
+#define CDC_ACM_CTRL_TRANSFER_SIZE (64)   // All standard CTRL requests and responses fit in this size
+#define CDC_ACM_CTRL_TIMEOUT_MS    (5000) // Every CDC device should be able to respond to CTRL transfer in 5 seconds
```
roma-jam, line 34: Seems, this parameter could be different, based on the device.... menuconfig configurable parameter?

roma-jam (approved):
Changed LGTM.

host/class/cdc/usb_host_cdc_acm/cdc_acm_host.c
tore-espressif: Yeah, I was considering it too. But because we don't have any menuconfig parameters for this driver and because we have not hit this timeout in the past 3 years, I wanted to leave it this way... 🤷

peter-marcisovsky (approved):
Changes LGTM.

PR #64, opened 2024-09-18
ci: Modernize esp_tinyusb test_app

tore-espressif:
We are preparing for more tests and test types (targets tests, CMock linux tests). Let's modernize our test_apps and pytest

* Add P4 to test_app build
* Use modern idf-build-app features
* Trim the build with `set(COMPONENTS main)` to get fast and lean build
* Upgrade tests for IDF v5.0 and above


roma-jam (approved):
Nothing to say, changes LGTM.

PR #68, opened 2024-09-30
UAC Class Host Driver v1.2.0

leeebo:
## Description

Release UAC Class Driver v1.2.0

### Background

> * For UAC devices, the correct way to configure the volume is first to obtain the maximum, minimum (a subset ranging from +127.9961 dB (0x7FFF) to -127.9961 dB (0x8001)), and step values. Then SET_CUR with the 16-bit db value (Please check UAC1.0 Spec Page.77 for details). v1.2.0 provides the standard implementation. 

### Breaking Changes:

1. Changed the parameter type of `uac_host_device_set_volume_db` from uint32_t to int16_t

### Improvements:

1. Support get current volume and mute status

### Bugfixes:

1. Fixed incorrect volume conversion. Using actual device volume range.
2. Fixed concurrency issues when suspend/stop during read/write

## Testing

Unitest Pass in both ESP32-S3 and ESP32-P4

---

## Checklist

### DIFF
--- a/host/class/uac/usb_host_uac/CHANGELOG.md
+++ b/host/class/uac/usb_host_uac/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog for USB Host UAC
 
+## 1.2.0 2024-09-27
+
+### Improvements:
+
+1. Support get current volume and mute status
+
+### Bugfixes:
+
+1. Fixed incorrect volume conversion. Using actual device volume range.
+2. Fixed concurrency issues when suspend/stop during read/write
+
 ## 1.1.0
 
 ### Improvements
--- a/host/class/uac/usb_host_uac/CHANGELOG.md
+++ b/host/class/uac/usb_host_uac/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## 1.2.0 2024-09-27
 
+### Breaking Changes:
+
+1. Changed the parameter type of `uac_host_device_set_volume_db` from uint32_t to int16_t
+
+
 ### Improvements:
 
 1. Support get current volume and mute status
--- a/host/class/uac/usb_host_uac/Kconfig
+++ b/host/class/uac/usb_host_uac/Kconfig
@@ -27,4 +27,9 @@ menu "USB Host UAC"
         default 3
         help
             Number of Packets per UAC ISOC URB. It limits the minimum packets each transfer will send.
+    config UAC_RINGBUF_SAFE_DELETE_WAITING_MS
+        int "Ringbuf Safe Delete Waiting Time in ms"
+        default 50
+        help
+            Ringbuf Safe Delay Time in ms. It is used to wait for the ringbuf to be untouched before deleting it.
 endmenu # "USB Host UAC"
--- a/host/class/uac/usb_host_uac/include/usb/uac_host.h
+++ b/host/class/uac/usb_host_uac/include/usb/uac_host.h
@@ -393,6 +393,19 @@ esp_err_t uac_host_device_write(uac_host_device_handle_t uac_dev_handle, uint8_t
 esp_err_t uac_host_device_set_mute(uac_host_device_handle_t uac_dev_handle, bool mute);
 
 /**
+ * @brief Get the mute status of the UAC device
+ * @param[in] iface       Pointer to UAC interface structure
+ * @param[out] mute       Pointer to store the mute status
+ * @return esp_err_t
+ * - ESP_OK on success
+ * - ESP_ERR_INVALID_STATE if the device is not ready or active
+ * - ESP_ERR_INVALID_ARG if the device handle is invalid
+ * - ESP_ERR_NOT_SUPPORTED if the device does not support mute control
+ * - ESP_ERR_TIMEOUT if the control timed out
+ */
+esp_err_t uac_host_device_get_mute(uac_host_device_handle_t uac_dev_handle, bool *mute);
+
+/**
  * @brief Set the volume of the UAC device
  * @param[in] iface       Pointer to UAC interface structure
  * @param[in] volume      Volume to set, 0-100
@@ -406,9 +419,23 @@ esp_err_t uac_host_device_set_mute(uac_host_device_handle_t uac_dev_handle, bool
 esp_err_t uac_host_device_set_volume(uac_host_device_handle_t uac_dev_handle, uint8_t volume);
 
 /**
+ * @brief Get the volume of the UAC device
+ * @param[in] iface       Pointer to UAC interface structure
+ * @param[out] volume     Pointer to store the volume, 0-100
+ * @return esp_err_t
+ * - ESP_OK on success
+ * - ESP_ERR_INVALID_STATE if the device is not ready or active
+ * - ESP_ERR_INVALID_ARG if the device handle is invalid
+ * - ESP_ERR_NOT_SUPPORTED if the device does not support volume control
+ * - ESP_ERR_TIMEOUT if the control timed out
+ */
+esp_err_t uac_host_device_get_volume(uac_host_device_handle_t uac_dev_handle, uint8_t *volume);
+
+/**
  * @brief Set the volume of the UAC device in dB
  * @param[in] iface       Pointer to UAC interface structure
- * @param[in] volume_db   Volume to set, db
+ * @param[in] volume_db   Volume to set, with resolution of 1/256 dB,
+ * eg.  0x0100 is 1 dB. 0x7FFF is 127.996 dB. 0x8001 is -127.996 dB.
  * @return esp_err_t
  * - ESP_OK on success
  * - ESP_ERR_INVALID_STATE if the device is not ready or active
@@ -417,6 +444,20 @@ esp_err_t uac_host_device_set_volume(uac_host_device_handle_t uac_dev_handle, ui
  * - ESP_ERR_TIMEOUT if the control timed out
  */
 esp_err_t uac_host_device_set_volume_db(uac_host_device_handle_t uac_dev_handle, uint32_t volume_db);
+
+/**
+ * @brief Get the volume of the UAC device in dB
+ * @param[in] iface       Pointer to UAC interface structure
+ * @param[out] volume_db  Pointer to store the volume, with resolution of 1/256 dB,
+ * eg.  0x0100 is 1 dB. 0x7FFF is 127.996 dB. 0x8001 is -127.996 dB.
+ * @return esp_err_t
+ * - ESP_OK on success
+ * - ESP_ERR_INVALID_STATE if the device is not ready or active
+ * - ESP_ERR_INVALID_ARG if the device handle is invalid
+ * - ESP_ERR_NOT_SUPPORTED if the device does not support volume control
+ * - ESP_ERR_TIMEOUT if the control timed out
+ */
+esp_err_t uac_host_device_get_volume_db(uac_host_device_handle_t uac_dev_handle, uint32_t *volume_db);
 
 #ifdef __cplusplus
 }
--- a/host/class/uac/usb_host_uac/include/usb/uac_host.h
+++ b/host/class/uac/usb_host_uac/include/usb/uac_host.h
@@ -381,7 +381,7 @@ esp_err_t uac_host_device_write(uac_host_device_handle_t uac_dev_handle, uint8_t
 
 /**
  * @brief Mute or un-mute the UAC device
- * @param[in] iface       Pointer to UAC interface structure
+ * @param[in] uac_dev_handle  UAC device handle
  * @param[in] mute        True to mute, false to unmute
  * @return esp_err_t
  * - ESP_OK on success
@@ -394,7 +394,7 @@ esp_err_t uac_host_device_set_mute(uac_host_device_handle_t uac_dev_handle, bool
 
 /**
  * @brief Get the mute status of the UAC device
- * @param[in] iface       Pointer to UAC interface structure
+ * @param[in] uac_dev_handle  UAC device handle
  * @param[out] mute       Pointer to store the mute status
  * @return esp_err_t
  * - ESP_OK on success
@@ -407,7 +407,7 @@ esp_err_t uac_host_device_get_mute(uac_host_device_handle_t uac_dev_handle, bool
 
 /**
  * @brief Set the volume of the UAC device
- * @param[in] iface       Pointer to UAC interface structure
+ * @param[in] uac_dev_handle  UAC device handle
  * @param[in] volume      Volume to set, 0-100
  * @return esp_err_t
  * - ESP_OK on success
@@ -420,7 +420,7 @@ esp_err_t uac_host_device_set_volume(uac_host_device_handle_t uac_dev_handle, ui
 
 /**
  * @brief Get the volume of the UAC device
- * @param[in] iface       Pointer to UAC interface structure
+ * @param[in] uac_dev_handle  UAC device handle
  * @param[out] volume     Pointer to store the volume, 0-100
  * @return esp_err_t
  * - ESP_OK on success
@@ -433,7 +433,7 @@ esp_err_t uac_host_device_get_volume(uac_host_device_handle_t uac_dev_handle, ui
 
 /**
  * @brief Set the volume of the UAC device in dB
- * @param[in] iface       Pointer to UAC interface structure
+ * @param[in] uac_dev_handle  UAC device handle
  * @param[in] volume_db   Volume to set, with resolution of 1/256 dB,
  * eg.  256 (0x0100) is 1 dB. 32767 (0x7FFF) is 127.996 dB. -32767 (0x8001) is -127.996 dB.
  * @return esp_err_t
@@ -447,7 +447,7 @@ esp_err_t uac_host_device_set_volume_db(uac_host_device_handle_t uac_dev_handle,
 
 /**
  * @brief Get the volume of the UAC device in dB
- * @param[in] iface       Pointer to UAC interface structure
+ * @param[in] uac_dev_handle  UAC device handle
  * @param[out] volume_db  Pointer to store the volume, with resolution of 1/256 dB,
  * eg.  256 (0x0100) is 1 dB. 32767 (0x7FFF) is 127.996 dB. -32767 (0x8001) is -127.996 dB.
  * @return esp_err_t
--- a/host/class/uac/usb_host_uac/include/usb/uac_host.h
+++ b/host/class/uac/usb_host_uac/include/usb/uac_host.h
@@ -435,7 +435,7 @@ esp_err_t uac_host_device_get_volume(uac_host_device_handle_t uac_dev_handle, ui
  * @brief Set the volume of the UAC device in dB
  * @param[in] iface       Pointer to UAC interface structure
  * @param[in] volume_db   Volume to set, with resolution of 1/256 dB,
- * eg.  0x0100 is 1 dB. 0x7FFF is 127.996 dB. 0x8001 is -127.996 dB.
+ * eg.  256 (0x0100) is 1 dB. 32767 (0x7FFF) is 127.996 dB. -32767 (0x8001) is -127.996 dB.
  * @return esp_err_t
  * - ESP_OK on success
  * - ESP_ERR_INVALID_STATE if the device is not ready or active
@@ -443,13 +443,13 @@ esp_err_t uac_host_device_get_volume(uac_host_device_handle_t uac_dev_handle, ui
  * - ESP_ERR_NOT_SUPPORTED if the device does not support volume control
  * - ESP_ERR_TIMEOUT if the control timed out
  */
-esp_err_t uac_host_device_set_volume_db(uac_host_device_handle_t uac_dev_handle, uint32_t volume_db);
+esp_err_t uac_host_device_set_volume_db(uac_host_device_handle_t uac_dev_handle, int16_t volume_db);
 
 /**
  * @brief Get the volume of the UAC device in dB
  * @param[in] iface       Pointer to UAC interface structure
  * @param[out] volume_db  Pointer to store the volume, with resolution of 1/256 dB,
- * eg.  0x0100 is 1 dB. 0x7FFF is 127.996 dB. 0x8001 is -127.996 dB.
+ * eg.  256 (0x0100) is 1 dB. 32767 (0x7FFF) is 127.996 dB. -32767 (0x8001) is -127.996 dB.
  * @return esp_err_t
  * - ESP_OK on success
  * - ESP_ERR_INVALID_STATE if the device is not ready or active
@@ -457,7 +457,7 @@ esp_err_t uac_host_device_set_volume_db(uac_host_device_handle_t uac_dev_handle,
  * - ESP_ERR_NOT_SUPPORTED if the device does not support volume control
  * - ESP_ERR_TIMEOUT if the control timed out
  */
-esp_err_t uac_host_device_get_volume_db(uac_host_device_handle_t uac_dev_handle, uint32_t *volume_db);
+esp_err_t uac_host_device_get_volume_db(uac_host_device_handle_t uac_dev_handle, int16_t *volume_db);
 
 #ifdef __cplusplus
 }
--- a/host/class/uac/usb_host_uac/test_app/main/test_host_uac.c
+++ b/host/class/uac/usb_host_uac/test_app/main/test_host_uac.c
@@ -619,6 +619,7 @@ TEST_CASE("test uac tx writing", "[uac_host][tx]")
     s_buffer += offset_size * freq_offsite_step;
 
     uint8_t volume = 0;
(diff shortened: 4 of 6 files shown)


leeebo:
@tore-espressif @roma-jam @peter-marcisovsky PTAL! Thanks1

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -157,6 +157,9 @@ typedef struct uac_interface {
     uint32_t ringbuf_size;                     /*!< Ring buffer size */
     uint32_t ringbuf_threshold;                /*!< Ring buffer threshold */
     uac_host_dev_info_t dev_info;              /*!< USB device parameters */
+    uint16_t vol_min_db;                       /*!< volume min with 1/256 db step */
+    uint16_t vol_max_db;                       /*!< volume max with 1/256 db step */
+    uint16_t vol_res_db;                       /*!< volume resolution with 1/256 db step */
```
tore-espressif, line 162: Is there any reason for using unsigned type for signed number?

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -2252,22 +2496,103 @@ esp_err_t uac_host_device_set_volume(uac_host_device_handle_t uac_dev_handle, ui
...
+
+    uac_host_interface_unlock(iface);
     return ESP_OK;
+
+fail:
+    uac_host_interface_unlock(iface);
+    return ret;
 }
 
 esp_err_t uac_host_device_set_volume_db(uac_host_device_handle_t uac_dev_handle, uint32_t volume_db)
```
peter-marcisovsky, line 2561: Do we need a 32 bit value for volume here? The `volume_db` is passed only to `uac_cs_request_set_volume()`, which expects `uint16_t` volume_db.
Same question for `uac_hos_device_get_volume_db`

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -2252,22 +2496,103 @@ esp_err_t uac_host_device_set_volume(uac_host_device_handle_t uac_dev_handle, ui
...
 {
     uac_iface_t *iface = get_iface_by_handle(uac_dev_handle);
     UAC_RETURN_ON_INVALID_ARG(iface);
```
peter-marcisovsky, line 2564: Should we also check for `volume_db` range here? Similarly to: 
`UAC_RETURN_ON_FALSE(volume <= 100, ESP_ERR_INVALID_ARG, "Invalid volume value");`
in `uac_host_device_set_volume()`

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -199,6 +202,34 @@ static esp_err_t _uac_host_device_delete(uac_device_t *uac_device);
...
+ * 0x0000: 0.0000db
+ * 0xFF00: -1.0000db (there is an typo in UAC 1.0 spec, it should be 0xFF00)
+ * 0xFE00: -2.0000db
+ * 0x8001: -127.9961db
+ * @param[in] volume      Volume in db
+ * @return uint16_t
+ */
+static float _volume_db_u16_2_f(uint16_t volume_db)
+{
+    return (volume_db & 0x8000 ? -128.0 : 0) + ((uint16_t)(volume_db & 0x7fff)) * 1.0 / 256;
```
peter-marcisovsky, line 219:
```suggestion
    return (volume_db & 0x8000 ? -128.0 : 0) + ((uint16_t)(volume_db & 0x7fff)) / 256.0f;
```
Can we do it like this? This way we will avoid multiplication by 1. 

Also, I am not sure if we have to explicitly set floating point numbers to float type, like this: `256.0f` or if the compiler would do it automatically.

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -2252,22 +2496,103 @@ esp_err_t uac_host_device_set_volume(uac_host_device_handle_t uac_dev_handle, ui
...
+        *volume = iface->cur_vol;
+        uac_host_interface_unlock(iface);
+        return ESP_OK;
+    }
+
+    // Otherwise, get the volume from the device
+    // Get volume range, calculate in dB float
+    uint16_t volume_db = 0;
+    UAC_GOTO_ON_ERROR(uac_cs_request_get_volume(iface, &volume_db), "Unable to get volume");
+    float volume_db_f = _volume_db_u16_2_f(volume_db);
```
peter-marcisovsky, line 2549:
```suggestion
    const float volume_db_f = _volume_db_u16_2_f(volume_db);
```

host/class/uac/usb_host_uac/Kconfig
```diff
@@ -27,4 +27,9 @@ menu "USB Host UAC"
         default 3
         help
             Number of Packets per UAC ISOC URB. It limits the minimum packets each transfer will send.
+    config UAC_RINGBUF_SAFE_DELETE_WAITING_MS
+        int "Ringbuf Safe Delete Waiting Time in ms"
+        default 50
```
peter-marcisovsky, line 32: Would it make sense to add some safe range for this value?

host/class/uac/usb_host_uac/test_app/main/test_host_uac.c
peter-marcisovsky, line 1: Could you consider adding test cases also for `uac_host_device_set_volume_db()` and `uac_host_device_get_volume_db()` functions?

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -199,6 +202,34 @@ static esp_err_t _uac_host_device_delete(uac_device_t *uac_device);
 static esp_err_t uac_cs_request_set(uac_device_t *uac_device, const uac_cs_request_t *req);
 static esp_err_t uac_cs_request_set_ep_frequency(uac_iface_t *iface, uint8_t ep_addr, uint32_t freq);
 
+// --------------------------- Utility Functions --------------------------------
+/**
+ * @brief Convert volume U16 type to UAC volume db, for details, please refer to UAC 1.0 spec P.77
+ * 0x7FFF: 127.9961db
+ * 0x0100: 1.0000db
+ * 0x0000: 0.0000db
+ * 0xFF00: -1.0000db (there is an typo in UAC 1.0 spec, it should be 0xFF00)
```
peter-marcisovsky, line 211:
```suggestion
 * 0xFF00: -1.0000db (there is a typo in UAC 1.0 spec, it should be 0xFF00)
```

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -2252,22 +2496,103 @@ esp_err_t uac_host_device_set_volume(uac_host_device_handle_t uac_dev_handle, ui
...
+    UAC_RETURN_ON_INVALID_ARG(iface);
+    UAC_RETURN_ON_INVALID_ARG(volume);
+    // Check if the device is active or ready
+    esp_err_t ret = ESP_OK;
+    UAC_RETURN_ON_ERROR(uac_host_interface_try_lock(iface, DEFAULT_CTRL_XFER_TIMEOUT_MS), "Unable to lock UAC Interface");
+    UAC_GOTO_ON_FALSE((UAC_INTERFACE_STATE_ACTIVE == iface->state || UAC_INTERFACE_STATE_READY == iface->state),
+                      ESP_ERR_INVALID_STATE, "device not ready or active");
+
+    // Return the backup volume value to avoid the volume reads differently than expected.
+    // Because the device volume adjustment step may relatively large,
```
peter-marcisovsky, line 2537:
```suggestion
    // Because the device volume adjustment step may be relatively large,
```

peter-marcisovsky (commented):
Leaving some minor comments

host/class/uac/usb_host_uac/uac_host.c
leeebo: @tore-espressif To be honest, this is because I mistakenly used the `uint32_t` type in `uac_host_device_set_volume_db` in the past, so I continued using unsigned type to maintain consistency in calculations.

leeebo: Or, how about making a "break change" to the `uac_host_device_set_volume_db` function, changing the param from uint32_t to int16_t? This would solve all the problems and make the calculations much simpler.

tore-espressif: I'd prefer introducing a breaking change and having a cleaner code :)

leeebo: @tore-espressif fixed in c2a3782b0f011edf2fb78814175ede1b035d27f8

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -109,6 +107,7 @@ typedef enum {
     UAC_INTERFACE_STATE_IDLE,                       /*!< UAC Interface has been opened but not started */
     UAC_INTERFACE_STATE_READY,                      /*!< UAC Interface has started but stream is suspended */
     UAC_INTERFACE_STATE_ACTIVE,                     /*!< UAC Interface is streaming */
+    UAC_INTERFACE_STATE_SUSPENDING,                 /*!< UAC Interface is suspending */
     UAC_INTERFACE_STATE_MAX
```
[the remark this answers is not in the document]
leeebo: The design is intended to be used for boundary checking, and since it is currentlly not used, let's delete it directly

host/class/uac/usb_host_uac/uac_host.c
leeebo: accept, thanks

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -1573,6 +1605,54 @@ static esp_err_t uac_cs_request_set(uac_device_t *uac_device, const uac_cs_reque
...
+    setup->bmRequestType = req->bmRequestType;
+    setup->bRequest = req->bRequest;
+    setup->wValue = req->wValue;
+    setup->wIndex = req->wIndex;
+    setup->wLength = req->wLength;
+
+    if (setup->bmRequestType == 0) {
+        setup->bmRequestType = USB_BM_REQUEST_TYPE_DIR_IN | USB_BM_REQUEST_TYPE_TYPE_CLASS
+                               | USB_BM_REQUEST_TYPE_RECIP_INTERFACE;
+    }
```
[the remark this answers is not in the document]
leeebo: Thanks for your advice, for better readability, let's break it up for now

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -1815,6 +2024,13 @@ esp_err_t uac_host_device_open(const uac_host_device_config_t *config, uac_host_
     UAC_ENTER_CRITICAL();
     uac_device->opened_cnt++;
     UAC_EXIT_CRITICAL();
+
+    // Get the current volume settings, we pass the error for the volume control may not be supported
```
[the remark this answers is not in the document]
leeebo: There are some UAC device not support volume control, please refer `4.3.2.5 Feature Unit Descriptor`

Let's make the code more clear, just call get_volume_range when it supported

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -1815,6 +2024,13 @@ esp_err_t uac_host_device_open(const uac_host_device_config_t *config, uac_host_
...
+    ret = uac_cs_request_get_volume_range(uac_iface, &uac_iface->vol_min_db, &uac_iface->vol_max_db, &uac_iface->vol_res_db);
+    if (ret != ESP_OK && ret != ESP_ERR_NOT_SUPPORTED) {
+        ESP_LOGW(TAG, "Failed to get volume range");
+    }
+
     return ESP_OK;
```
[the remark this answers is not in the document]
leeebo: fixed

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -2239,11 +2455,39 @@ esp_err_t uac_host_device_set_mute(uac_host_device_handle_t uac_dev_handle, bool
...
 
-    UAC_RETURN_ON_ERROR(uac_cs_request_set_mute(iface, mute), "Unable to set mute");
+    UAC_GOTO_ON_ERROR(uac_cs_request_set_mute(iface, mute), "Unable to set mute");
+    ESP_LOGI(TAG, "%s Interface %d-%d", mute ? "Mute" : "Unmute", iface->dev_info.iface_num, iface->cur_alt + 1);
+    uac_host_interface_unlock(iface);
     return ESP_OK;
+
+fail:
+    uac_host_interface_unlock(iface);
+    return ret;
```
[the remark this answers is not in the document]
leeebo: the ret is accessed in `UAC_GOTO_ON_ERROR`

host/class/uac/usb_host_uac/uac_host.c
leeebo: Yes, you are right, let's check it below

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -2239,11 +2455,39 @@ esp_err_t uac_host_device_set_mute(uac_host_device_handle_t uac_dev_handle, bool
...
+    UAC_GOTO_ON_FALSE((UAC_INTERFACE_STATE_ACTIVE == iface->state || UAC_INTERFACE_STATE_READY == iface->state),
+                      ESP_ERR_INVALID_STATE, "device not ready or active");
+
+    UAC_GOTO_ON_ERROR(uac_cs_request_get_mute(iface, mute), "Unable to get mute");
+    uac_host_interface_unlock(iface);
+    return ESP_OK;
+
+fail:
+    uac_host_interface_unlock(iface);
+    return ret;
```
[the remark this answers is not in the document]
leeebo: same as above

host/class/uac/usb_host_uac/uac_host.c
leeebo: fixed now, using standard int16_t instead

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -199,6 +202,34 @@ static esp_err_t _uac_host_device_delete(uac_device_t *uac_device);
...
+/**
+ * @brief Convert volume U16 type to UAC volume db, for details, please refer to UAC 1.0 spec P.77
+ * 0x7FFF: 127.9961db
+ * 0x0100: 1.0000db
+ * 0x0000: 0.0000db
+ * 0xFF00: -1.0000db (there is an typo in UAC 1.0 spec, it should be 0xFF00)
+ * 0xFE00: -2.0000db
+ * 0x8001: -127.9961db
+ * @param[in] volume      Volume in db
+ * @return uint16_t
```
roma-jam, line 214:
```suggestion
 * @return float
```

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -199,6 +202,34 @@ static esp_err_t _uac_host_device_delete(uac_device_t *uac_device);
...
+ */
+static float _volume_db_u16_2_f(uint16_t volume_db)
+{
+    return (volume_db & 0x8000 ? -128.0 : 0) + ((uint16_t)(volume_db & 0x7fff)) * 1.0 / 256;
+}
+
+static uint16_t _volume_db_f_2_u16(float volume_db_f)
+{
+    uint16_t volume_db = 0;
```
roma-jam: When we have several arithmetic operation in sequence, sometimes it is better to split them logically and apply masks (if necessary) with specific flags. Also, it helps to verify the edge or overflow cases. 

Something like: 

```suggestion
    bool negative = false;
    if(volume_db_f < 0) {
        volume_db_f += 128.0; // could here be an overflow ? 
        negative = true;
    }
    
    volume_db = (uint16_t)round(volume_db_f * 256); // could here be an overflow? 
    
    return (negative)? volume_db | 0x8000 : volume_db;
    
```

roma-jam (commented):
Generally LGTM, just couple of suggestions.

host/class/uac/usb_host_uac/uac_host.c
roma-jam: Not applied anymore

roma-jam: Oh, no, still valid!

host/class/uac/usb_host_uac/Kconfig
leeebo: TBH, adding waiting before deleting is just a workaround. I am looking for a safe way to delete the ring buffer in a multithreaded environment. Any suggestions?

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -2239,11 +2455,39 @@ esp_err_t uac_host_device_set_mute(uac_host_device_handle_t uac_dev_handle, bool
...
-    UAC_RETURN_ON_ERROR(uac_cs_request_set_mute(iface, mute), "Unable to set mute");
+    UAC_GOTO_ON_ERROR(uac_cs_request_set_mute(iface, mute), "Unable to set mute");
+    ESP_LOGI(TAG, "%s Interface %d-%d", mute ? "Mute" : "Unmute", iface->dev_info.iface_num, iface->cur_alt + 1);
+    uac_host_interface_unlock(iface);
     return ESP_OK;
+
+fail:
+    uac_host_interface_unlock(iface);
+    return ret;
+}
```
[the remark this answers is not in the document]
leeebo: I prefer the current for better readable

host/class/uac/usb_host_uac/uac_host.c
leeebo: @roma-jam Thanks for your advice. I have changed to using int16_t to avoid manual manipulation of negative numbers. c2a3782b0f011edf2fb78814175ede1b035d27f8

Add I also add the check to avoid overflow fc3ca5d53d49b100f6f2ffe3402f9c4968b1da4a

leeebo: The value 127.9961 is from the UAC 1.0 spec

image

host/class/uac/usb_host_uac/include/usb/uac_host.h
```diff
@@ -392,6 +392,19 @@ esp_err_t uac_host_device_write(uac_host_device_handle_t uac_dev_handle, uint8_t
  */
 esp_err_t uac_host_device_set_mute(uac_host_device_handle_t uac_dev_handle, bool mute);
 
+/**
+ * @brief Get the mute status of the UAC device
+ * @param[in] iface       Pointer to UAC interface structure
```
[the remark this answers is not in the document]
leeebo: Fixed, Thanks!

host/class/uac/usb_host_uac/CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog for USB Host UAC
 
+## 1.2.0 2024-09-27
```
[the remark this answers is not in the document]
leeebo: thanks, fixed

host/class/uac/usb_host_uac/include/usb/uac_host.h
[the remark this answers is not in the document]
leeebo: @roma-jam Finally, I keep the name  `uac_dev_handle`, because `dev_hdl` is used as `usb_device_handle_t`

host/class/uac/usb_host_uac/Kconfig
roma-jam: @leeebo , I think the answer here is pretty standard: everything, that needs to be closed only after releasing should track the opening status. So, that could be flag or counter to active clients (pretty much as for usb devices in our library).
You claim/lock the device (ring buffer) when it is opened. When the device is gone, you check the amount of clients, and if there is some, you notify them, that this device should be closed (ring buffer released). During release the ring buffer you decrease the counter, and when it is 0 you can delete the ring buffer itself.

tore-espressif (approved):
LGTM!

host/class/uac/usb_host_uac/Kconfig
leeebo: @roma-jam Thanks for your suggestion. I realized using the counter to prevent ring buffer from being deleted when used, but this requires introducing another mutex lock to protect the counter, because the read/write API may pass a large wait delay (hold the lock for a long time)

roma-jam: @leeebo yes, sure it is. Because here we are start to talking about members protection in multitask systems. 
But I think that might be done as a separate PR, as the current one already contain a lot of changes.

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -199,6 +201,33 @@ static esp_err_t _uac_host_device_delete(uac_device_t *uac_device);
...
+ * @return float
+ */
+static float _volume_db_i16_2_f(int16_t volume_db)
+{
+    return (float)(volume_db) / 256.0f;
+}
+
+static int16_t _volume_db_f_2_i16(float volume_db_f)
+{
+    if (volume_db_f > 127.9961) {
```
roma-jam, line 223: Usually, it is a good idea to make all of the constants as a `#define` variable. 
In that case, if you need (just as example) to change the 127.9961 in the code, you will have only one place where you need to do that: `#define VOLUME_MAX_DB (127.9961f)`

roma-jam, line 223: In the `_volume_db_i16_2_f` you have the `f` next to the float data. 
Maybe it is better to keep the same style everywhere. 

```suggestion
    if (volume_db_f > 127.9961f) {
```

roma-jam (approved):
Two small nitpicks. 
Otherwise, really huge amount of work! Thank you so much!

host/class/uac/usb_host_uac/test_app/main/test_host_uac.c
leeebo: added

host/class/uac/usb_host_uac/uac_host.c
leeebo: @peter-marcisovsky the 256 will be implicitly converted to 256.0f

 for your reference: https://en.cppreference.com/w/c/language/conversion

image

leeebo: To avoid confusion, I've added an `f` suffix to all decimals

peter-marcisovsky: Thx for the reference. Looks good.

host/class/uac/usb_host_uac/uac_host.c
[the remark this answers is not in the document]
peter-marcisovsky: Please correct me if I am wrong, but from the deffinition:

Looks like the `UAC_GOTO_ON_ERROR` does not accept any extra `esp_err_t ret` parameter. It works with the one which is returned from a function with is being checked for a return value.
`#define UAC_GOTO_ON_ERROR(exp, msg) ESP_GOTO_ON_ERROR(exp, fail, TAG, msg)`

Like here for example: 
`UAC_GOTO_ON_ERROR(uac_cs_request_set_mute(iface, mute), "Unable to set mute");` 
The `ret` is not passed directly to the `UAC_GOTO_ON_ERROR`, but the function `uac_cs_...` provide a ret value directly.


While for example the `UAC_GOTO_ON_FALSE` does accept an `esp_err_t ret` value
`#define UAC_GOTO_ON_FALSE(exp, err, msg) ESP_GOTO_ON_FALSE((exp), err, fail, TAG, msg)`

For example here:
`UAC_GOTO_ON_FALSE(uac_iface->state_mutex, ESP_ERR_NO_MEM, "Unable to create state mutex");`
Which we can also do like: 
`UAC_GOTO_ON_FALSE(uac_iface->state_mutex, ret, "Unable to create state mutex");`

Or is it other way around? 
I mean, it's not that important right now, but just for my own interest. Thx for an explanation.

peter-marcisovsky (approved):
LGTM, thanks for the changes.

host/class/uac/usb_host_uac/uac_host.c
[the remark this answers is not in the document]
tore-espressif: @peter-marcisovsky This is defined in `esp_check.h`

``` c
#define ESP_GOTO_ON_ERROR(x, goto_tag, log_tag, format, ...) do {                           \
    esp_err_t err_rc_ = (x);                                                                \
    if (unlikely(err_rc_ != ESP_OK)) {                                                      \
        ESP_LOGE(log_tag, "%s(%d): " format, __FUNCTION__, __LINE__, ##__VA_ARGS__);        \
        ret = err_rc_;            <------ 'ret' accessed here
        goto goto_tag;                                                                      \
    }                                                                                       \
} while(0)
```

So this macro assumes that a variable `ret` exists outside of its context... It is not clear indeed, but this is how we use it.

I'll resolve this conversation so we can go ahead with merging, we can discuss offline, if needed :)

roma-jam:
@leeebo don‘t forget to squash before merge ;)

PR #71, opened 2024-10-02
Add option to open a CDC device with any VID or PID

tore-espressif:
## Description
The CDC driver must support also devices who do not follow the CDC specification. Any many cases, the user does not know or care about the device's VID/PID. With this feautere, any device can be opened

## Testing

A test is added that checks all positive (device opened) and negative (device not opened) cases

## Checklist

### DIFF
--- a/host/class/cdc/usb_host_cdc_acm/CHANGELOG.md
+++ b/host/class/cdc/usb_host_cdc_acm/CHANGELOG.md
@@ -1,7 +1,8 @@
-## [Unreleased]
+## 2.0.5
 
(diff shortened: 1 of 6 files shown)


peter-marcisovsky (approved):
Changes LGTM, 

I tried `cdc_acm_host` example on this branch. The `CDC_HOST_ANY_` Works fine. 

For some devices the example fails, but it's just an incorrect device being plugged in and has nothing to do with this changes, for example it fails with the Silicon Labs USB UART Bridge present on our dev boards. 

With External hubs, the example crashes all the time as expected. So we could warn users about that, also in the example.

PR #75, opened 2024-10-08
ci(esp_tinyusb): Run esp_tinyusb test apps in CI

peter-marcisovsky:
## Description
This MR enables `esp_tinyusb` tests in CI on a new `usb_device` runner

## Related
IDF-11345

## Testing
To pass the CI

---

## Checklist

Before submitting a Pull Request, please ensure the following:

- [x] 🚨 This PR does not introduce breaking changes.
- [ ] All CI checks (GH Actions) pass.
- [ ] Documentation is updated as needed.
- [ ] Tests are updated or added as necessary.
- [x] Code is well-commented, especially in complex areas.
- [ ] Git history is clean — commits are squashed to the minimum necessary.

PR #86, opened 2024-11-15
feat(ci): Build and run target tests everyday

tore-espressif:
In https://github.com/espressif/esp-usb/pull/75 we refactored the test runners' configuration. We want to run the test more frequently so we can evaluate their stability.

This entire commit can be reverted after the evaluation period is over (in couple of weeks).


tomassebestik (approved):
LGTM!

PR #93, opened 2024-11-28
change(test_vendor): Compatibility with tinyusb v0.17.x and higher 

roma-jam:
## Description
USB Device Vendor specific class driver API changed from v0.16.0 to v0.17.0.

`void tud_vendor_rx_cb(uint8_t itf)` - > `void tud_vendor_rx_cb(uint8_t itf, uint8_t const *buffer, uint16_t bufsize)`

To build `test_apps` from esp_tinyusb component with tinyusb component, with v0.17.0.1 or higher, the test_app code should be changed accordingly. 

## Related

- Caused by `espressif/tinyusb` [v0.17.0~1](https://github.com/espressif/tinyusb/tree/v0.17.0.1)

## Testing

---

## Checklist

### DIFF
--- a/.github/workflows/build_idf_examples.yml
+++ b/.github/workflows/build_idf_examples.yml
@@ -24,5 +24,5 @@ jobs:
           pip install idf-component-manager==1.5.2 idf-build-apps==2.4.3 --upgrade
           python .github/ci/override_managed_component.py esp_tinyusb device/esp_tinyusb ${IDF_PATH}/examples/peripherals/usb/device/tusb_*
           cd ${IDF_PATH}
-          idf-build-apps find --path examples/peripherals/usb/device/ --recursive --target all --manifest-file examples/peripherals/.build-test-rules.yml
-          idf-build-apps build --path examples/peripherals/usb/device/ --recursive --target all --manifest-file examples/peripherals/.build-test-rules.yml
+          idf-build-apps find --path examples/peripherals/usb/device/ --recursive --target all --manifest-file examples/peripherals/.build-test-rules.yml --build-dir build_@t_@w --work-dir @f_@t_@w
+          idf-build-apps build --path examples/peripherals/usb/device/ --recursive --target all --manifest-file examples/peripherals/.build-test-rules.yml --build-dir build_@t_@w --work-dir @f_@t_@w
--- a/device/esp_tinyusb/test_apps/vendor/main/test_vendor.c
+++ b/device/esp_tinyusb/test_apps/vendor/main/test_vendor.c
@@ -21,7 +21,11 @@
 static const char *TAG = "vendor_test";
 
 char buffer_in[64];
+#if (TUSB_VERSION_MINOR >= 17)
+void tud_vendor_rx_cb(uint8_t itf, uint8_t const *buffer, uint16_t bufsize)
+#else
 void tud_vendor_rx_cb(uint8_t itf)
+#endif // TUSB_VERSION_MINOR
 {
     ESP_LOGI(TAG, "tud_vendor_rx_cb(itf=%d)", itf);
     int available = tud_vendor_n_available(itf);


tore-espressif:
@roma-jam There is something wrong with the CI job, I'll take a look

roma-jam:
@tore-espressif, yes we saw that with Peter yesterday. 
I don’t know why after build first example for p4, the component manager thinks that tinyusb has changed on disk. 

I gave tried to solve it with changing component/idf-build-apps versions, but seems that that is not the reason.

So I fixed the broken vendor test_app and leave it like this. 

P.S. What kind of magic spell did you exactly do with the path? )

tore-espressif:
@roma-jam The job passed now.

Previously, we build 1 example for 3 targets but all in the same `build` and example folder. This probably caused conflict during target switch... (although it should not)

You can check `build-dir` and `work-dir` arguments for you reference here https://docs.espressif.com/projects/idf-build-apps/en/latest/references/cli.html#named-arguments

We should also double check that TinyUSB does not generate any files during CMake configuration phase to the managed_componets folder...

device/esp_tinyusb/test_apps/vendor/main/test_vendor.c
```diff
@@ -21,7 +21,11 @@
 static const char *TAG = "vendor_test";
 
 char buffer_in[64];
+#if (TUSB_VERSION_MINOR >= 17)
```
peter-marcisovsky, line 24: Are the this changes even needed? Updating build dir and work dir seems to fix the problem.

.github/workflows/build_idf_examples.yml
```diff
@@ -24,5 +24,5 @@ jobs:
           pip install idf-component-manager==1.5.2 idf-build-apps==2.4.3 --upgrade
           python .github/ci/override_managed_component.py esp_tinyusb device/esp_tinyusb ${IDF_PATH}/examples/peripherals/usb/device/tusb_*
           cd ${IDF_PATH}
-          idf-build-apps find --path examples/peripherals/usb/device/ --recursive --target all --manifest-file examples/peripherals/.build-test-rules.yml
-          idf-build-apps build --path examples/peripherals/usb/device/ --recursive --target all --manifest-file examples/peripherals/.build-test-rules.yml
+          idf-build-apps find --path examples/peripherals/usb/device/ --recursive --target all --manifest-file examples/peripherals/.build-test-rules.yml --build-dir build_@t_@w --work-dir @f_@t_@w
```
peter-marcisovsky, line 27: I started to see this problem (when IDF example is built in the same directory for different targets in CI) yesterday. Did something change with the component manager? Or why this has not been an issue until now?

tore-espressif: We first want to fix it here so we have passing CI in esp-usb.

We use fixed versions of idf-component-manager and idf-build-apps. So the only change can be in esp-idf itself

peter-marcisovsky:
@roma-jam here is an [example](https://docs.espressif.com/projects/idf-build-apps/en/latest/references/config_file.html#usage) from `idf-build-apps`

device/esp_tinyusb/test_apps/vendor/main/test_vendor.c
roma-jam: Yes, they remove another problem and unlock CI for esp-usb.

tore-espressif:
> I’ll need an explanation for this kind of spell: `--build-dir build_@t_@w --work-dir @f_@t_@w` 🙂

The placeholders are descibed in first paragraph here https://docs.espressif.com/projects/idf-build-apps/en/latest/references/cli.html#cli-reference

Here you can see how the placeholders are expanded https://github.com/espressif/esp-usb/actions/runs/12082344074/job/33693227937?pr=93#step:4:86

The '@w' placeholder is not used (just a copy paste from different repo)

tore-espressif (approved):

PR #96, opened 2024-12-03
fix(esp_tinyusb): Pinned tinyusb task to the CPU1 by default

roma-jam:
# [esp_tinyusb component](https://components.espressif.com/components/espressif/esp_tinyusb), v1.5.0

## Requirements

A problem could be present in a high-loaded systems (with several interfaces) and when large amounts of data were being sent and received simultaneously, the dcd_event_xfer_complete event was not triggered. 

## Description

Pinned tinyusb FreeRTOS task to CPU1 by default in the dual-core systems. 

## Related

- Releated to IDF-4552

## Testing
N/A

## Checklist

### DIFF
--- a/device/esp_tinyusb/Kconfig
+++ b/device/esp_tinyusb/Kconfig
@@ -55,6 +55,7 @@ menu "TinyUSB Stack"
 
         choice TINYUSB_TASK_AFFINITY
             prompt "TinyUSB task affinity"
+            default TINYUSB_TASK_AFFINITY_CPU1 if !FREERTOS_UNICORE
             default TINYUSB_TASK_AFFINITY_NO_AFFINITY
             depends on !TINYUSB_NO_DEFAULT_TASK
             help
--- a/device/esp_tinyusb/CHANGELOG.md
+++ b/device/esp_tinyusb/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 1.5.0 (Unreleased)
+## 1.5.0
 
 - esp_tinyusb: Added DMA mode option to tinyusb DCD DWC2 configuration 
+- esp_tinyusb: Changed the default affinity mask of the task to CPU1
 
 ## 1.4.5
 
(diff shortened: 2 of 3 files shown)


tore-espressif:
@roma-jam
* please add the change to CHANGELOG.mg
* Make sure it works as expected in single core system (ESP32-S2)

roma-jam:
@tore-espressif If we have the CI such as: Build and Run USB Test Application (which builds and runs the test_apps on esp32s2), does it counts that it works "as expected"? 

Or you are talking about something else?

tore-espressif:
> does it counts that it works "as expected"?

Yes, that is actually enough :D Maybe just add a comment that it is OK top pin to CPU1 on single CPU system. The users might get confused

device/esp_tinyusb/Kconfig
```diff
@@ -41,7 +41,7 @@ menu "TinyUSB Stack"
 
         choice TINYUSB_TASK_AFFINITY
             prompt "TinyUSB task affinity"
-            default TINYUSB_TASK_AFFINITY_NO_AFFINITY
+            default TINYUSB_TASK_AFFINITY_CPU1
             depends on !TINYUSB_NO_DEFAULT_TASK
             help
```
tore-espressif, line 46: This is what I can see now for ESP32-S2

image

Selecting task affininty on single-core system might confuse the users (Why should I choose CPU affinity if there is only 1 CPU?). 

Could you please fix this so the 'TinyUSB task affinity' menu `depends on !FREERTOS_UNICORE` ?

roma-jam: I hate the invert logic (especially, `bool "Do not create a TinyUSB task"` part, which led in the code to `#if !CONFIG_TINYUSB_INIT_IN_DEFAULT_TASK`). I want to refactor everything. But later. 

I created a ticket for that and right now just made a dependency of default parameter from `FREERTOS_UNICORE` not to change the config too much.  Because for `esp32s2` the AFFINITY_MASK is absolutely pointless.
:angry:

tore-espressif (approved):

peter-marcisovsky (approved):

PR #100, opened 2024-12-06
fix(ci): Enable IDF 5.3 test app run in CI for esp32p4

peter-marcisovsky:
## Description

This MR Enables IDF 5.3 test app run in CI for esp32p4, as we were waiting for the docker image of the IDF 5.3 to update. 
This update in IDF 5.3 adds support for esp32p4 ECO2 flashing.

## Testing

To pass this CI

## Checklist


roma-jam (approved):

roma-jam:
One nitpick about artifacts, we don't need **/test_app*/**/build_esp*/test_app_*.elf for the pytest.
Usually, *.elf is used for debugging, so we don't need to pass it as an artifact.

peter-marcisovsky:
@roma-jam thx for checking, we have been passing elf flies as an artifact since the very beginning. But, yes, looks like we can remove it.

tore-espressif:
> @roma-jam thx for checking, we have been passing elf flies as an artifact since the very beginning. But, yes, looks like we can remove it.

BTW, the .elf file is actually used by pytest in case there is 'abort' or similar failure. It will use the elf file to decode at which line the program aborted

peter-marcisovsky:
@roma-jam looks like we should better leave the .elf files in the artifacts. I will revert the latest changes

PR #101, opened 2024-12-09
feat(msc_host): Add format MSC support

igi540:
## Requirements
Provide the possibility to format connected flash drive

## Description
Currently the flash drive should be formatted in advance, otherwise it can be formatted by ESP when mount error is present. To have the possibility to format connected flash drives in the USB Host MSC class driver explicitly, the new API call has been added. Now, the format process of the flash drive can be done not only during the mounting failure (format_if_mount_failed and FR_NO_FILESYSTEM) but explicitly from the application logic.
## Related
- Closes https://github.com/espressif/esp-idf/issues/14890
- IDFGH-14076 Missing option for USB MSC drive format
## Testing
Added test case for testing public API.

---

## Checklist

### DIFF
--- a/host/class/msc/usb_host_msc/include/usb/msc_host_vfs.h
+++ b/host/class/msc/usb_host_msc/include/usb/msc_host_vfs.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2023 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2024 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -15,6 +15,20 @@ extern "C" {
 #endif
 
 typedef struct msc_host_vfs *msc_host_vfs_handle_t;           /**< VFS handle to attached Mass Storage device */
+
+/**
+ * @brief Format MSC device.
+ *
+ * @param[in] device Device handle obtained from MSC callback provided upon initialization
+ * @param[in] mount_config Mount configuration
+ * @param[in] vfs_handle Handle to MSC device associated with registered VFS
+ * @return esp_err_t
+ * @return
+ *    - ESP_OK: Format completed
+ *    - ESP_ERR_INVALID_ARG: All arguments must be present and couldn't be NULL
+ *    - ESP_ERR_MSC_FORMAT_FAILED: Formatting failed
+ */
+esp_err_t msc_host_vfs_format(msc_host_device_handle_t device, const esp_vfs_fat_mount_config_t *mount_config, const msc_host_vfs_handle_t vfs_handle);
 
 /**
  * @brief Register MSC device to Virtual filesystem.
--- a/host/class/msc/usb_host_msc/CHANGELOG.md
+++ b/host/class/msc/usb_host_msc/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [Unreleased] 
+
+- Added public API support for formatting
+
 ## 1.1.3 
 
 - Implemented request sense, to get sense data from USB device in case of an error
--- a/host/class/msc/usb_host_msc/src/msc_host_vfs.c
+++ b/host/class/msc/usb_host_msc/src/msc_host_vfs.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2023 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2024 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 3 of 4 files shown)


host/class/msc/usb_host_msc/include/usb/msc_host_vfs.h
```diff
@@ -16,6 +16,16 @@ extern "C" {
...
 
+/**
+ * @brief Format MSC device.
+ *
+ * @param[in] device Device handle obtained from MSC callback provided upon initialization
+ * @param[in] mount_config Mount configuration
+ * @param[in] vfs_handle Handle to MSC device associated with registered VFS
+ * @return esp_err_t
+ */
+esp_err_t esp_vfs_fat_usb_msc_format(msc_host_device_handle_t device, const esp_vfs_fat_mount_config_t *mount_config, const msc_host_vfs_handle_t vfs_handle);
```
roma-jam, line 27: Usually, all the function names have the same pattern inside the one file. 
Here, we have two public calls: 
- `msc_host_vfs_register`
- `msc_host_vfs_unregister`

So if we want to provide a new API within current module/files, I would assume that it should be something like: 
```suggestion
esp_err_t msc_host_vfs_format(msc_host_device_handle_t device, const esp_vfs_fat_mount_config_t *mount_config, const msc_host_vfs_handle_t vfs_handle);
```

host/class/msc/usb_host_msc/include/usb/msc_host_vfs.h
```diff
@@ -16,6 +16,16 @@ extern "C" {
 
 typedef struct msc_host_vfs *msc_host_vfs_handle_t;           /**< VFS handle to attached Mass Storage device */
 
+/**
+ * @brief Format MSC device.
+ *
+ * @param[in] device Device handle obtained from MSC callback provided upon initialization
+ * @param[in] mount_config Mount configuration
+ * @param[in] vfs_handle Handle to MSC device associated with registered VFS
+ * @return esp_err_t
```
roma-jam, line 25: Here we usually describe the possible return values. The description for the errors should be clear, when you read them. So, for example, for error argument: 
- "Device handle could not be NULL" is better than just the "Wrong argument".

```suggestion
 * @return
 *    - ESP_OK: Format completed  
 *    - ESP_ERR_INVALID_ARG: All arguments must be present and couldn't be NULL
```

roma-jam (approved):
Just two small nitpicks, otherwise LGTM.

host/class/msc/usb_host_msc/include/usb/msc_host_vfs.h
tore-espressif: @roma-jam The naming was actually my idea, I wanted to make it similar to other formatting functions that we have in esp-idf, eg sdcard_format

https://github.com/espressif/esp-idf/blob/a2f420a36ecf8cd19b443adde1d75e5eaecc6309/components/fatfs/vfs/esp_vfs_fat.h#L268

I'd like to rename `msc_host_vfs_register/unregister` accordingly too... WDYT?

roma-jam: @tore-espressif Oh, I see. 

For `esp_vfs_fat` it makes sense, as the component itself has the similar name. 

If apply the same logic on our case .. it should be `esp_usb_msc_host_format` or something like that.

peter-marcisovsky (approved):
We can also update our tests, and use this new function.

[Here](https://github.com/espressif/esp-usb/blob/master/host/class/msc/usb_host_msc/test_app/main/test_msc.c#L440), we have test case called `can_be_formatted` where we force fail mounting of the partition, to trigger the formatting.

Thus, we could either use this new public API function, to format the partition in this existing test case, instead of force failing the mounting. 

Or to create a new, similar test case, where we create a file, format the partition and check if the file exists.

Otherwise LGTM

host/class/msc/usb_host_msc/include/usb/msc_host_vfs.h
tore-espressif: OK, let's go for consistency within the module, so `msc_host_vfs_format`

host/class/msc/usb_host_msc/include/usb/msc_host_vfs.h
igi540: thank you for suggestion, fixed.

igi540:
@peter-marcisovsky ok, thank you for suggesting about adding test case. I added another test case. While updating test I found a [line](https://github.com/espressif/esp-usb/blob/master/host/class/msc/usb_host_msc/test_app/main/test_msc.c#L451) which I am not sure whether it is correct. Why does it erase memory before formatting? PTAL

tore-espressif (approved):
Thank you @igi540 !
Let's have a look at https://github.com/espressif/esp-usb/blob/e5feb3fabef0c64ffc054fe3eb4c9cf52c9fcdb3/host/class/msc/usb_host_msc/test_app/main/test_msc.c#L451 in follow-up work

Issue #107, opened 2024-12-21
USB-NCM driver  0.17 hangs after a period of sending packets  (IEC-243)


emaayan:
### Answers checklist.

- [X] I have read the component documentation [ESP-IDF Components](https://components.espressif.com) and the issue is not addressed there.
- [X] I am using target and esp-idf version as defined in component's idf_component.yml
- [X] I have searched the [issue tracker](https://github.com/espressif/esp-usb/issues?q=is%3Aissue) for a similar issue and not found any related issue.

### Which component are you using? If you choose Other, provide details in More Information.

device/esp_tinyusb

### ESP-IDF version.

5.3.1

### Development Kit.

esp-s3

### Used Component version.

0.17.1

### More Information.

hi, i'm directly using tinyusb port from espressif , but that repo has no issues repo 

if i try to use the esp_tinyusb component wrapper the tinyusb_send method would eventually fail consistently with sending this. 
steps to reproduce: 
1.  download and install the windows ncm driver 
2. upload the sketch 
3.  when the network device is added give it a static IP 192.168.5.3 , and gateway 192.168.5.1 
[esp_tusb_ncm_bug.zip](https://github.com/user-attachments/files/18219870/esp_tusb_ncm_bug.zip)

the program generates a bytes array at random lengths from 20 to 1400 and first sends the size of the byte array and the actuall byte array while the corresponding client code does the opposite 

after a while it just halts in it tracks sometimes you get log error of ethernetif_input: IP input error
` 
private static ByteBuffer getByteBuffer(InputStream inputStream, int size) throws IOException {
        final ByteBuffer allocate = ByteBuffer.allocate(size).order(ByteOrder.LITTLE_ENDIAN);
        final byte[] array = allocate.array();
        final int read = inputStream.readNBytes(array,0,size);// read(array,0,size);
        return allocate;
    }
    public static void main(String[] args) throws IOException {
         String host = "192.168.5.1";
        final Socket socket = SocketFactory.getDefault().createSocket(host, 19000);
        while (socket.isConnected()){
            final InputStream inputStream = socket.getInputStream();
            final ByteBuffer byteBuffer = getByteBuffer(inputStream, 4);
            final int sz = byteBuffer.getInt();
            final byte[] bytes = inputStream.readNBytes(sz);
            System.out.println(sz +" " +Arrays.toString(bytes));
        }
    }
`

it hangs on tud_task() , which waits indefinitely..

roma-jam:
[...]

Issue #46, opened 2024-07-21
Extremely slow RNDIS throughput with esp-tinyusb 


wuyuanyi135:
### Answers checklist.

- [X] I have read the component documentation [ESP-IDF Components](https://components.espressif.com) and the issue is not addressed there.
- [X] I am using target and esp-idf version as defined in component's idf_component.yml
- [X] I have searched the [issue tracker](https://github.com/espressif/esp-usb/issues?q=is%3Aissue) for a similar issue and not found any related issue.

### Which component are you using? If you choose Other, provide details in More Information.

device/esp_tinyusb

### ESP-IDF version.

v5.2

### Development Kit.

ESP32-S3-WROOM-1-N8R8

### Used Component version.

1.4.4 (Dirty)

### More Information.

In my project, a USB dongle made of ESP32-S3 are used as communication gateway that creates an RNDIS interface on the PC for connecting with the other devices.

### Current setup
I have been using the tinyusb and additions from [esp-iot-bridge](https://github.com/espressif/esp-iot-bridge/tree/master/components/usb/usb_device) and the throughtput was satisfactory (max ~7Mbps). Because of the USB refactoring work, the esp-iot-bridge solution does not compile on esp-idf > 5.2. 

### Code difference
The iot-bridge's `[tinyusb_install_driver](https://github.com/espressif/esp-iot-bridge/blob/master/components/usb/usb_device/additions/src/tinyusb.c)` uses tons of hal code but the [`esp-tinyusb` counterpart ](https://github.com/espressif/esp-usb/blob/master/device/esp_tinyusb/tinyusb.c) uses new API that is compatible with `esp-idf>5.2`.

However, from my observation, with esp_tinyusb, the throughtput was only about 400 Kbps, which is much slower than what I could get from esp-iot-bridge. Looking at the implementation of ECM_RNDIS in tinyusb, it seems there isn't any significant change from the versions. 

### Benchmark

| idf     | esp-tinyusb    | tinyusb                                                                                    | net   | speed (Mbps) | comment                                                                                          |
|---------|----------------|--------------------------------------------------------------------------------------------|-------|--------------|--------------------------------------------------------------------------------------------------|
| 5.2     | esp-iot-bridge | 0.11.0                                                                                     | RNDIS |      ~7      | baseline                                                                                         |
| 5.2     | esp-tinyusb    | 0.15.0~10                                                                                  | RNDIS |     ~0.4     |                                                                                                  |
[...]

PR #114, opened 2025-01-09
fix(esp_tinyusb): Fix NCM NTB buffers configuration

### LINKED ISSUE #113: Number of NTB buffers configuration option
### Is your feature request related to a problem?

With the current NTB buffer settings, USB-NCM does not seem to work properly and does not achieve optimal performance.
In my setup the following error occurs quite frequently:
tud_network_can_xmit: request blocked

### Describe the solution you'd like.

Add configuration option for CFG_TUD_NCM_OUT_NTB_N and CFG_TUD_NCM_IN_NTB_N via component config.
I.e. enhancing tusb_config.h of the component with:
#define CFG_TUD_NCM_IN_NTB_N CONFIG_TINYUSB_NCM_IN_NTB_N
#define CFG_TUD_NCM_OUT_NTB_N CONFIG_TINYUSB_NCM_OUT_NTB_N

and adding to KConfig relevant lines to configure it.

### Describe alternatives you've considered.

_No response_

### Additional context.

_No response_

peter-marcisovsky:
## Description

This MR adds a possibility to configure NCM transfer blocks via menuconfig.

A table below shows RAM usage and performance dependency on NTB buffers configuration, NTB buffers count and NTB buffers length. Memory occupation was measured using [NCM Device Example](https://github.com/espressif/esp-idf/tree/master/examples/peripherals/usb/device/tusb_ncm) on esp32s3.

| NTB Buffers count (IN/OUT) | 1/1             | 2/2             | 2/2           | 3/3             | 4/4            |
|--------------------------|-----------------|-----------------|---------------|-----------------|----------------|
| NTB Buffers length (IN/OUT)   | 3200/3200       | 3200/3200       | 6400/6400     | 8192/8192       | 8192/8192      |
| Used RAM [bytes]         | 112320          | 118736          | 131552        | 155120          | 171520         |
| Used RAM [%]             | 32.9            | 34.7            | 38.5          | 45.5            | 50.2           |
| Download [Mbps]          | 1.19            | 4.22            | 5.7           | 7.21            | 7.6            |
| Upload [Mbps]            | 5.04            | 5.67            | 5.96          | 6.14            | 6.03           |

Measuring the performance did prove a significant gain dependency on NTB buffers configurations, similarly to the description in [tinyusb](https://github.com/espressif/tinyusb/blob/37995c70595164aa82f5b13496ae71c9584224b9/src/class/net/ncm.h#L61l) upstream is caliming for RP2040.

NTB Buffers count 1/1 and length 3200/3200 is a default value in tinyusb upstream.
The most efficient settings (RAM to performance gain ratio) seems to be the 4th option (NTB Buffers count 3/3, Buffers length 8192/8192).

Increasing the NTB Buffers count and length did also help to mitigate the [tud_network_can_xmit: request blocked](https://github.com/espressif/tinyusb/blob/37995c70595164aa82f5b13496ae71c9584224b9/src/class/net/ncm.h#L63) warning messages while running NCM device.

## Related

Closes #113 
Relates to #107 

## Testing

Performance measurement using [Iperf](https://github.com/espressif/esp-idf/blob/master/examples/wifi/iperf/README.md) did not prove to be useful.

Online speed test was used for the performance measurement (using the esp32s3 NCM device WIFi station to wired interface)

Below is a performance comparison between the 
 - tinyusb's upstream default settings of the NTB buffers (count 1/1, length 3200/3200)
 - NTB Buffers configured to: count 4/4, length 8192/8192

## NTB Buffers(IN/OUT) count: 1/1, length(bytes): 3200/3200 

alt text

## NTB Buffers(IN/OUT) count: 4/4, length(bytes): 8192/8192 

alt text

---

## Checklist

### DIFF
--- a/device/esp_tinyusb/CHANGELOG.md
+++ b/device/esp_tinyusb/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [Unreleased]
+
+- esp_tinyusb: Added possibility to configure NCM Transfer Blocks (NTB) via menuconfig
+
 ## 1.6.0
 
 - CDC-ACM: Fixed memory leak on deinit
--- a/device/esp_tinyusb/Kconfig
+++ b/device/esp_tinyusb/Kconfig
@@ -285,6 +285,51 @@ menu "TinyUSB Stack"
             config TINYUSB_NET_MODE_NONE
                 bool "None"
         endchoice
+
+        config TINYUSB_NCM_OUT_NTB_BUFFS_COUNT
+            int "Number of NCM NTB buffers for reception side"
+            depends on TINYUSB_NET_MODE_NCM
+            default 3
+            range 1 6
+            help
+                Number of NTB buffers for reception side.
+                Can be increased to improve performance and stability with the cost of additional RAM requirements.
+                Helps to mitigate "tud_network_can_xmit: request blocked" warning message when running NCM device.
+
+        config TINYUSB_NCM_IN_NTB_BUFFS_COUNT
+            int "Number of NCM NTB buffers for transmission side"
+            depends on TINYUSB_NET_MODE_NCM
+            default 3
+            range 1 6
+            help
+                Number of NTB buffers for transmission side.
+                Can be increased to improve performance and stability with the cost of additional RAM requirements.
+                Helps to mitigate "tud_network_can_xmit: request blocked" warning message when running NCM device.
+
+        config TINYUSB_NCM_OUT_NTB_BUFF_MAX_SIZE
+            int "NCM NTB Buffer size for reception size"
+            depends on TINYUSB_NET_MODE_NCM
+            default 8192
+            range 1600 10240
+            help
+              Size of NTB buffers on the reception side. The minimum size used by Linux is 2048 bytes.
+              NTB buffer size must be significantly larger than the MTU (Maximum Transmission Unit).
+              The typical default MTU size for Ethernet is 1500 bytes, plus an additional packet overhead.
+              To improve performance, the NTB buffer size should be large enough to fit multiple MTU-sized
+              frames in a single NTB buffer and it's length should be multiple of 4.
+
+        config TINYUSB_NCM_IN_NTB_BUFF_MAX_SIZE
+            int "NCM NTB Buffer size for transmission size"
+            depends on TINYUSB_NET_MODE_NCM
+            default 8192
+            range 1600 10240
+            help
+              Size of NTB buffers on the transmission side. The minimum size used by Linux is 2048 bytes.
+              NTB buffer size must be significantly larger than the MTU (Maximum Transmission Unit).
+              The typical default MTU size for Ethernet is 1500 bytes, plus an additional packet overhead.
+              To improve performance, the NTB buffer size should be large enough to fit multiple MTU-sized
+              frames in a single NTB buffer and it's length should be multiple of 4.
+
     endmenu # "Network driver (ECM/NCM/RNDIS)"
 
     menu "Vendor Specific Interface"
--- a/device/esp_tinyusb/include/tusb_config.h
+++ b/device/esp_tinyusb/include/tusb_config.h
@@ -1,6 +1,6 @@
 /*
  * SPDX-FileCopyrightText: 2019 Ha Thach (tinyusb.org),
- * SPDX-FileContributor: 2020-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileContributor: 2020-2025 Espressif Systems (Shanghai) CO LTD
  * SPDX-License-Identifier: MIT
  *
  * Copyright (c) 2019 Ha Thach (tinyusb.org),
@@ -172,6 +172,12 @@ extern "C" {
 #define CFG_TUD_DFU                 CONFIG_TINYUSB_DFU_MODE_DFU
 #define CFG_TUD_DFU_RUNTIME         CONFIG_TINYUSB_DFU_MODE_DFU_RUNTIME
 #define CFG_TUD_BTH                 CONFIG_TINYUSB_BTH_ENABLED
+
+// NCM NET Mode NTB buffers configuration
+#define CFG_TUD_NCM_OUT_NTB_N         CONFIG_TINYUSB_NCM_OUT_NTB_BUFFS_COUNT
+#define CFG_TUD_NCM_IN_NTB_N          CONFIG_TINYUSB_NCM_IN_NTB_BUFFS_COUNT
+#define CFG_TUD_NCM_OUT_NTB_MAX_SIZE  CONFIG_TINYUSB_NCM_OUT_NTB_BUFF_MAX_SIZE
+#define CFG_TUD_NCM_IN_NTB_MAX_SIZE   CONFIG_TINYUSB_NCM_IN_NTB_BUFF_MAX_SIZE
 
 #ifdef __cplusplus
 }


device/esp_tinyusb/Kconfig
```diff
@@ -285,6 +285,41 @@ menu "TinyUSB Stack"
...
+            default 2
+            range 1 4
+            help
+                Number of NTB buffers for reception side.
+                Can be increased to improve performance with the cost of additional RAM requirements.
+
+        config TINYUSB_NCM_IN_NTB_BUFFS_COUNT
+            int "Number of NCM NTB buffers for transmission side"
+            depends on TINYUSB_NET_MODE_NCM
+            default 2
```
roma-jam, line 301: IMHO it is better to use the best default value, which is 3.

https://github.com/espressif/tinyusb/blob/37995c70595164aa82f5b13496ae71c9584224b9/src/class/net/ncm.h#L63

```suggestion
            default 3
```

device/esp_tinyusb/Kconfig
```diff
@@ -285,6 +285,41 @@ menu "TinyUSB Stack"
...
+            range 1 4
+            help
+                Number of NTB buffers for transmission side.
+                Can be increased to improve performance with the cost of additional RAM requirements.
```
roma-jam, line 305: Maybe it is better to provide some details, why do we need to change this config. As we don't have any tool to measure the numbers so far, maybe we can provide only general idea: 

```suggestion
                Can be increased to improve performance with the cost of additional RAM requirements.
                Possible options: 
                1 - Good performance, but some packet could be lost
                2 - Better performance on small packets, but sometimes transfer request could be blocked with an error
                 3 - Better performance, transfer request can never be blocked
                 4 - No additional gain
```

device/esp_tinyusb/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [Unreleased]
+
+- esp_tinyusb: Added configuration for NCM device NTB buffers
```
roma-jam, line 3:
```suggestion
- esp_tinyusb: Added possibility to configure NCM Transfer Blocks (NTB) via menuconfig
```

roma-jam (commented):
Two general ideas: 
1. Default value for `CFG_TUD_NCM_IN_NTB_N`: 2 or 3? 
2. Based on the NCM specification the NTB - is a `NCM Transfer Block`. So the shortend `NCM NTB` means` NCM NCM Transfer Block` which might be not necessary. 
Maybe, just check the menuconfig description based on this thought.

device/esp_tinyusb/Kconfig
peter-marcisovsky: The suggestions from tinyusb are based on `RPi RP2040`.  And for the [STM32](https://github.com/espressif/tinyusb/blob/37995c70595164aa82f5b13496ae71c9584224b9/src/class/net/ncm.h#L66) for example, there is no performance gain at all.

I did some Iperf testing, but there was no such performance gain in 50%, or 30%, as the description in tinyusb for RP2040 is claiming.

Also the statement [3 - "tud_network_can_xmit: request blocked" never happens](https://github.com/espressif/tinyusb/blob/37995c70595164aa82f5b13496ae71c9584224b9/src/class/net/ncm.h#L63) is not true, in our case. I had to set the buffer count really high (to around 10) to completely mitigate the warning message.

Sure, we can have the default value 3, but for esp32, we can't really say that there is some visible performance gain, or `request blocked` will never happen.

device/esp_tinyusb/Kconfig
peter-marcisovsky: I was not able to prove any of those statements, so I bit rephrased the description. But yes, I can put some guidelines there.

device/esp_tinyusb/Kconfig
roma-jam: Oh, cool.
Then, if we have the real test results, why we can't use them? 
That means, that for our case, to exclude the blocking, we need configure it to 10. 

BTW, is that the same for Slave and DMA mode?

device/esp_tinyusb/Kconfig
roma-jam: I didn't know that you have tested it. 
If these statements are not relevant, then there is no point to change anything.

device/esp_tinyusb/Kconfig
peter-marcisovsky: Bit difficult to say what is the optimal count here.
 
Wen running IPERF, setting the NTB Count to 2 or 3 was enough, to get the NCM stable (no `request blocked` happened at all).

But then, I used esp32s3 as a network device [sta2eth](https://github.com/espressif/esp-idf/tree/master/examples/network/sta2eth) example or [tusb_ncm](https://github.com/espressif/esp-idf/tree/master/examples/peripherals/usb/device/tusb_ncm) example. So practically I used esp32s3 connected to a local wifi network, forwarding connection to my PC over USB for casual Internet browsing (google maps, youtube... ) In this case, I had to set the buff count to 10, to stop seeing `request blocked`. And this looks like a bit of extreme scenario to me.

The DMA settings had no effect on this if I remember. I can check again.

peter-marcisovsky:
@roma-jam  @tore-espressif  PTAL at the performance results from IPERF.

I did not observe any significant performance when configuring the NCM Device settings, but as stated in my [comment](https://github.com/espressif/esp-usb/pull/114#discussion_r1910088324), I still observed somer reliability issues when running some higher traffic through the device, whit fairly "low" count of the NTB buffs.

I am thinking, what would be a reasonable setting of ranges in Kconfig for the NTB buffs count.

roma-jam (approved):
Thanks for the testing and providing iperf results! 
LGTM

P.S. Do not forget to squash before merge.

device/esp_tinyusb/Kconfig
tore-espressif: > Good performance, but some packets could be lost

@roma-jam What do you mean by 'lost packets' here? The esp_netif will handle the packet retransmission, and as far as I know, this is considered normal in networking. 'lost packet' could scare some users about data being entirely lost...

tore-espressif:
@peter-marcisovsky @roma-jam Thank you for all the testing. I also run some tests (with Win10), here are my results:

1. Our WiFi (20-30Mbs/) is faster than USB FS (12Mb/s), so some requests will get 'blocked' just because we cannot send them so fast through USB... This is generally not a problem, the packet will be retransmitted by the networking layer
2. Default Ethernet MTU size is 1500bytes (plus we need some more for metadata). Thus, increasing the buffer length from 3200 too 4096 bytes will have almost no effect, as both sizes can fit only 2 datagrams. Increasing it to eg. 8096 will have great effect because now we can send 5 datagrams in one NTB buffer! -> I'd propose increasing the max size in menconfig to 16kB, just to give the users enough flexibility
3. Funny things is, that with 4 NTBs and 8196 buffers, I got 8Mbit/s. I think we this is very close to Bulk FS throughput limit -> we should investigate why are users reporting extremely slow throughput https://github.com/espressif/esp-usb/issues/46
4. I was also able to reproduce https://github.com/espressif/esp-usb/issues/107 pretty often, seems to be unrelated to what you are fixing in this PR
5. Please also check https://github.com/espressif/tinyusb/pull/48 . It should fix some ugly bugs in the NCM driver

device/esp_tinyusb/Kconfig
roma-jam: @tore-espressif lost events.

Lost packets sounds scaring, indeed.

peter-marcisovsky:
> 1. Our WiFi (20-30Mbs/) is faster than USB FS (12Mb/s), so some requests will get 'blocked' just because we cannot send them so fast through USB... This is generally not a problem, the packet will be retransmitted by the networking layer

I am getting `request blocked` on Linux even with 4 buffs and 8192 buff lengths, but I wouldn't call it an issue either. Did you get rid of the warning message completely on WIndows?

> 2. Default Ethernet MTU size is 1500bytes (plus we need some more for metadata). Thus, increasing the buffer length from 3200 to 4096 bytes will have almost no effect, as both sizes can fit only 2 datagrams. Increasing it to eg. 8096 will have great effect because now we can send 5 datagrams in one NTB buffer! -> I'd propose increasing the max size in menconfig to 16kB, just to give the users enough flexibility

Good findings. 
Iperf did not get better even after increasing the lenghts to 8192, but I was running Iperf between 2 ESPs, so there could be something else.. or maybe the Iperf settings.

Running speed tests looks like a better idea, I did not think about it. On Linux I am getting `7.5 Mbps Download` and `6.1 Mbps Upload` with 4 Buffs 8192 lengths.. Maybe bit worse connection in the office? Eitherway, nice improvement comparing to the TInyUSB's default buffs settings where I got only `1.5 Mbps download`.

Going to 16384 lengths was for some reason so bad. I was getting huge latency. Can you try 16k buff lengths as well?

> 3. Funny things is, that with 4 NTBs and 8196 buffers, I got 8Mbit/s. I think this is very close to Bulk FS throughput limit -> we should investigate why are users reporting extremely slow throughput [Extremely slow RNDIS throughput with esp-tinyusb  ([IEC-137](https://jira.espressif.com:8443/browse/IEC-137)) #46](https://github.com/espressif/esp-usb/issues/46)

I think, those buffer settings we are configuring in this MR, are valid only for NCM device, not for RNDIS device. Maybe there is some different setting for RNDIS which we can configure.

@tore-espressif  thank you for the investigations, and idea of using Speed test (so obvious..) I will update the MR based on this findings.

tore-espressif:
>  Did you get rid of the warning message completely on WIndows?

No no, I still got the message (if I enabled it :D ). But it is expected behavior. USB is bottleneck in this data stream

> On Linux I am getting 7.5 Mbps Download and 6.1 Mbps Upload with 4 Buffs 8192 lengths.. Maybe bit worse connection in the office?

I doubt it would be so bad in the office. 7.5Mbps is good enough for Full Speed bulk transfer

> Going to 16384 lengths was for some reason so bad. 

I did not test, will do

> are valid only for NCM device, not for RNDIS device

NCM is mentioned in https://github.com/espressif/esp-usb/issues/46 too... in the 'Benchmark' table

tore-espressif:
@peter-marcisovsky I tested again, and it seems that TinyUSB driver breaks if the buffer are > (10*1024) bytes.

So I'd limit max buffer length to 10240.

device/esp_tinyusb/Kconfig
```diff
@@ -285,6 +285,51 @@ menu "TinyUSB Stack"
...
+            int "NCM NTB Buffer size for reception size"
+            depends on TINYUSB_NET_MODE_NCM
+            default 8192
+            range 1600 10000
+            help
+              Size of NTB buffers on the reception side. The minimum size used by Linux is 2048 bytes.
+              NTB buffer size must be significantly larger than the MTU (Maximum Transmission Unit).
+              The typical default MTU size for Ethernet is 1500 bytes, plus an additional packet overhead.
+              To improve performance, the NTB buffer size should be large enough to fit multiple MTU-sized
+              frames in a single NTB buffer.
```
tore-espressif, line 319: Also, the size must be aligned to 4. (Divisible by 4)

peter-marcisovsky: Fixed. Added a note about it.

tore-espressif (approved):

PR #84, opened 2024-11-05
USB Host UVC driver version 2

tore-espressif:
## Introduction

This PR introduces version 2 of the USB Host UVC driver, aimed at enabling ESP SoCs to stream frames from connected USB cameras.

The driver provides frames to the user via a callback, offering a flexible interface that can serve as a foundation for more advanced frame processing applications.

Internally, the driver manages a FreeRTOS Queue containing multiple frame buffers, which helps prevent buffer overflows and underflows during streaming.

Detailed feature descriptions, future plans, and usage instructions are available in the README.md file.

## Information for Testers
This section provides instructions for testing the driver’s functionality and public API.

Two examples are included for testing:

### 1. `basic_uvc_stream`
   - **Test Setup**: Tested on ESP32-P4 with a customer-specified dual-camera setup.
   - **Functionality**: Opens two UVC streams and continuously starts and stops them.

### 2. `camera_display`
   - **Test Setup**: Tested on ESP32-S3 with multiple compatible cameras.
   - **Functionality**: Initializes a display, opens a single UVC stream, decodes received JPEG frames, and displays them on the screen.

## Information for Reviewers
This section provides guidance for those reviewing the internal code structure of the driver.

The driver includes a `host_test` folder with tests designed to run on Linux. These tests use a mocked version of the esp-idf/usb component, focusing on the following areas, which should need less in-depth review:
* **Configuration descriptor parsing** – `uvc_descriptor_parsing.c`
* **USB transfer callbacks** – `uvc_isoc.c` and `uvc_bulk.c`

The following file, focused on human-readable output, also requires only a light review:
* **Descriptor printing** – `uvc_descriptor_printing.c`

Areas not yet covered by `host_tests` and requiring more detailed review include:
* **Frame buffer management** – `uvc_frame.c` (manages queue of empty frame buffers)
* **Control requests** – `uvc_control.c` (handles video format negotiation and other control requests)
* **Core driver functionality** – `uvc_host.c` (includes driver installation, device management, stream initiation, etc.)

## Related Information
* This PR supersedes #60.

PR #129, opened 2025-01-30
fix(host/uvc): Fixed negotiation for some non-conforming UVC devices

tore-espressif:
Some UVC devices require to set alternate setting to 0 before the negotiation starts.

Closes https://github.com/espressif/esp-idf/issues/9868

### DIFF
--- a/host/class/uvc/usb_host_uvc/uvc_control.c
+++ b/host/class/uvc/usb_host_uvc/uvc_control.c
@@ -1,11 +1,12 @@
 /*
- * SPDX-FileCopyrightText: 2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2024-2025 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
 // This file will contain all Class-Specific request from USB UVC specification chapter 4
 
 #include <string.h> // For memset
+#include <math.h>   // fabsf for float comparison
 
 #include "esp_check.h"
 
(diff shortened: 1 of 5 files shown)


tore-espressif:
cc @lijunru-hub

roma-jam:
Still have the same problem as I described here: https://github.com/espressif/esp-usb/pull/84#issuecomment-2538713360
This is the diff I made to run the `example/basic_uvc_stream`: 
image

This is the log: 
image


What may be related or can give you some ideas (but I need to change it everytime): on my board the ESP PSRAM uses Octal Mode.

roma-jam (approved):
Only one note, mentioned above. 
Otherwise, LGTM.

host/class/uvc/usb_host_uvc/uvc_control.c
```diff
@@ -16,6 +17,8 @@
 #include "uvc_descriptors_priv.h"
 #include "uvc_check_priv.h"
 
+#define FLOAT_EQUAL(a, b) (fabs(a - b) < 0.0001f) // For comparing float values with acceptable difference (epsilon value)
```
peter-marcisovsky, line 20: Here, you are using math function for `double` type, not for `float`
```suggestion
#define FLOAT_EQUAL(a, b) (fabsf(a - b) < 0.0001f) // For comparing float values with acceptable difference (epsilon value)
```

peter-marcisovsky (approved):
Just one note. Tired on my Logitech camera, works fine.

tore-espressif:
The 1 remaining issue is solved here https://github.com/espressif/esp-usb/pull/153

PR #131, opened 2025-02-05
fix(tusb_msc): Cache sector metrics, enable storage buffering

igi540:
### 🚀 **Optimize USB MSC Performance on ESP32-S3, ESP32-P4, and ESP32-S2**

---

## ✅ **Requirements**

**Why is this needed?**  
- **ESP32-S3 USB Mass Storage (MSC) implementation showed suboptimal performance** (slow file copy speed).  
- **Observed issues:**
  - ⚠️ **SPI Flash write speed was severely limited** due to architectural constraints.  
  - 📉 **SD card speeds varied significantly based on FIFO size**, but were still below expectations.  
- **Goal:** Optimize data transfer efficiency for both SPI Flash and SD cards.

---

## 📝 **Description**

This PR improves USB Mass Storage (MSC) performance by:
- ✅ **Caching `get_sector_size` and `get_sector_count`** to reduce redundant queries and improve efficiency.
- ✅ **Optimizing buffering logic:**
  - **A single-buffer approach is used for all targets** (ESP32-S3, ESP32-P4, ESP32-S2)
  - **Buffering significantly improves SD card write speed**, while SPI Flash shows minimal improvement due to hardware constraints.

---

#### **SD Card Performance**
| FIFO Size | Read Speed (Old) | Read Speed (New) | Write Speed (Old) | Write Speed (New) |
|-----------|------------------|------------------|-------------------|-------------------|
| **512B**  | 0.561 MB/s       | 0.566 MB/s       | 0.191 MB/s        | 0.236 MB/s        |
| **8192B** | 0.925 MB/s       | 0.925 MB/s       | 0.726 MB/s        | 0.928 MB/s    |

✅ **Write speed improved significantly with 8192B FIFO (~27% increase).**  
✅ **Read speed remains stable at ~0.9 MB/s.**

#### **SPI Flash Performance**
| FIFO Size | Write Speed (Old) | Write Speed (New) |
|-----------|------------------|------------------|
| **512B**  | 8.21 KB/s         | 8.46 KB/s        |
| **8192B** | 30.31 KB/s        | 30.2 KB/s        |

⚠ **Minimal improvement for SPI Flash (limited by write speed).**

---

#### **SD Card Performance**
| FIFO Size | Read Speed (Old) | Read Speed (New) | Write Speed (Old) | Write Speed (New) |
|-----------|------------------|------------------|-------------------|-------------------|
| **512B**  | 1.175 MB/s       | 1.174 MB/s       | 0.234 MB/s        | 0.238 MB/s        |
| **8192B** | 4.786 MB/s       | 4.744 MB/s       | 2.021 MB/s        | 2.157 MB/s        |
| **32768B**| 5.833 MB/s       | **5.998 MB/s**   | 3.673 MB/s        | **4.485 MB/s**    |

✅ **Larger FIFO sizes (32KB) significantly improved SD card write speed (~22% increase).**

#### **SPI Flash Performance**
| FIFO Size | Write Speed (Old) | Write Speed (New) |
|-----------|------------------|------------------|
| **512B**  | 7.44 KB/s         | 7.57 KB/s        |
| **8192B** | 28.44 KB/s        | 28.42 KB/s       |
| **32768B**| 28.12 KB/s        | 28.24 KB/s       |

🔸 **No major changes in SPI Flash performance on ESP32-P4.**

---

#### **SPI Flash Performance (Corrected Values)**
| FIFO Size | Write Speed (Old) | Write Speed (New) |
|-----------|------------------|------------------|
| **512B**  | 5.50 KB/s     | 5.59 KB/s    |
| **8192B** | 21.52 KB/s        | 21.54 KB/s       |

🔸 **Minimal improvements observed, similar to ESP32-S3 results.**  
⚠ **SPI Flash performance remains limited due to architectural constraints.**

🚫 **Note:** SD card is **currently not supported in MSC device mode** for ESP32-S2.

---

## 🧪 **Measurement Method**

- **SD card performance** was measured using **SpeedOut** on Windows (applies to all supported **SoCs**).
- **SPI Flash performance** was measured using a **custom Windows Bash script** (applies to all supported **SoCs**).

---

## 🚩 **Why SPI Flash Performance is Limited?**

The **core limitation** is that **program execution and storage reside on the same flash chip**:  
- 🧠 **During a flash write operation, program execution from flash is suspended.**  
- ⏱ **Interrupts not loaded into IRAM cannot run**, severely slowing down USB transfers.  
- ⚠️ This explains **why SPI Flash remains slow**, even with buffering optimizations.

**SD cards, on the other hand, are not affected by this limitation,** which is why they show significant improvements with optimized buffering.

---

## 🔗 **Related**

- Closes [GitHub Issue #11110](https://github.com/espressif/esp-idf/issues/11110)

---

## 📌 **Conclusion**

✅ **SD card performance now approaches USB Full-Speed theoretical limit (~1 MB/s on ESP32-S3).**  
✅ **Write speed improvements are significant for SD cards, especially with larger FIFO buffers (up to ~4.5 MB/s on ESP32-P4).**  
⚠ **SPI Flash remains speed-limited due to architectural constraints, with only minor improvements observed.**

🚀 **This PR significantly enhances SD card performance and optimizes MSC efficiency across ESP32 targets!**

### DIFF
--- a/device/esp_tinyusb/tusb_msc_storage.c
+++ b/device/esp_tinyusb/tusb_msc_storage.c
@@ -13,41 +13,69 @@
 #include "diskio_wl.h"
 #include "wear_levelling.h"
 #include "esp_partition.h"
+#include "esp_memory_utils.h"
+#include "sdkconfig.h"
 #include "vfs_fat_internal.h"
 #include "tinyusb.h"
+#include "device/usbd_pvt.h"
 #include "class/msc/msc_device.h"
 #include "tusb_msc_storage.h"
-#include "esp_vfs_fat.h"
 #if SOC_SDMMC_HOST_SUPPORTED
 #include "diskio_sdmmc.h"
 #endif
 
 static const char *TAG = "tinyusb_msc_storage";
 
+#define MSC_STORAGE_MEM_ALIGN 4
+#define MSC_STORAGE_BUFFER_SIZE CONFIG_TINYUSB_MSC_BUFSIZE /*!< Size of the buffer, configured via menuconfig (MSC FIFO size) */
+
+#if ((MSC_STORAGE_BUFFER_SIZE) % MSC_STORAGE_MEM_ALIGN != 0)
+#error "CONFIG_TINYUSB_MSC_BUFSIZE must be divisible by MSC_STORAGE_MEM_ALIGN. Adjust your configuration (MSC FIFO size) in menuconfig."
+#endif
+
 // CONFIG_TINYUSB_MSC_BUFSIZE must be bigger, or equal to the CONFIG_WL_SECTOR_SIZE
 #if (CONFIG_TINYUSB_MSC_BUFSIZE < CONFIG_WL_SECTOR_SIZE)
 #error "CONFIG_TINYUSB_MSC_BUFSIZE must be at least at the size of CONFIG_WL_SECTOR_SIZE"
 #endif
 
+/**
+ * @brief Structure representing a single write buffer for MSC operations.
+ */
 typedef struct {
-    bool is_fat_mounted;
-    const char *base_path;
+    uint8_t data_buffer[MSC_STORAGE_BUFFER_SIZE]; /*!< Buffer to store write data. The size is defined by MSC_STORAGE_BUFFER_SIZE. */
+    uint32_t lba;                          /*!< Logical Block Address for the current WRITE10 operation. */
+    uint32_t offset;                       /*!< Offset within the specified LBA for the current write operation. */
+    uint32_t bufsize;                      /*!< Number of bytes to be written in this operation. */
+} msc_storage_buffer_t;
+
+/**
+ * @brief Handle for TinyUSB MSC storage interface.
+ *
+ * This structure holds metadata and function pointers required to
+ * manage the underlying storage medium (SPI flash, SDMMC).
+ */
+typedef struct {
+    msc_storage_buffer_t storage_buffer;
+    bool is_fat_mounted;                  /*!< Indicates if the FAT filesystem is currently mounted. */
+    const char *base_path;                /*!< Base path where the filesystem is mounted. */
     union {
-        wl_handle_t wl_handle;
+        wl_handle_t wl_handle;            /*!< Handle for wear leveling on SPI flash. */
 #if SOC_SDMMC_HOST_SUPPORTED
-        sdmmc_card_t *card;
+        sdmmc_card_t *card;               /*!< Handle for SDMMC card. */
 #endif
     };
-    esp_err_t (*mount)(BYTE pdrv);
-    esp_err_t (*unmount)(void);
-    uint32_t (*sector_count)(void);
-    uint32_t (*sector_size)(void);
-    esp_err_t (*read)(size_t sector_size, uint32_t lba, uint32_t offset, size_t size, void *dest);
-    esp_err_t (*write)(size_t sector_size, size_t addr, uint32_t lba, uint32_t offset, size_t size, const void *src);
-    tusb_msc_callback_t callback_mount_changed;
-    tusb_msc_callback_t callback_premount_changed;
-    int max_files;
-} tinyusb_msc_storage_handle_s; /*!< MSC object */
+    esp_err_t (*mount)(BYTE pdrv);        /*!< Pointer to the mount function. */
+    esp_err_t (*unmount)(void);           /*!< Pointer to the unmount function. */
+    uint32_t sector_count;                /*!< Total number of sectors in the storage medium. */
+    uint32_t sector_size;                 /*!< Size of a single sector in bytes. */
+    esp_err_t (*read)(size_t sector_size, /*!< Function pointer for reading data. */
+                      uint32_t lba, uint32_t offset, size_t size, void *dest);
+    esp_err_t (*write)(size_t sector_size, /*!< Function pointer for writing data. */
+                       size_t addr, uint32_t lba, uint32_t offset, size_t size, const void *src);
+    tusb_msc_callback_t callback_mount_changed; /*!< Callback for mount state change. */
+    tusb_msc_callback_t callback_premount_changed; /*!< Callback for pre-mount state change. */
+    int max_files;                          /*!< Maximum number of files that can be open simultaneously. */
+} tinyusb_msc_storage_handle_s;
 
 /* handle of tinyusb driver connected to application */
 static tinyusb_msc_storage_handle_s *s_storage_handle;
@@ -175,7 +203,7 @@ static esp_err_t _write_sector_sdmmc(size_t sector_size,
 }
 #endif
 
-static esp_err_t msc_storage_read_sector(uint32_t lba,
+static esp_err_t _msc_storage_read_sector(uint32_t lba,
         uint32_t offset,
         size_t size,
         void *dest)
@@ -185,7 +213,7 @@ static esp_err_t msc_storage_read_sector(uint32_t lba,
     return (s_storage_handle->read)(sector_size, lba, offset, size, dest);
 }
 
-static esp_err_t msc_storage_write_sector(uint32_t lba,
+static esp_err_t _msc_storage_write_sector(uint32_t lba,
         uint32_t offset,
         size_t size,
         const void *src)
@@ -252,6 +280,29 @@ fail:
         free(workbuf);
     }
     return ret;
+}
+
+/**
+ * @brief Handles deferred USB MSC write operations.
+ *
+ * This function is invoked via TinyUSB's deferred execution mechanism to perform
+ * write operations to the underlying storage. It writes data from the
+ * `storage_buffer` stored within the `s_storage_handle`.
+ *
+ * @param param Unused. Present for compatibility with deferred function signature.
+ */
+static void _write_func(void *param)
+{
+    // Process the data in storage_buffer
+    esp_err_t err = _msc_storage_write_sector(
+                        s_storage_handle->storage_buffer.lba,
+                        s_storage_handle->storage_buffer.offset,
+                        s_storage_handle->storage_buffer.bufsize,
+                        (const void *)s_storage_handle->storage_buffer.data_buffer
+                    );
+    if (err != ESP_OK) {
+        ESP_LOGE(TAG, "Write failed, error=0x%x", err);
+    }
 }
 
 esp_err_t tinyusb_msc_storage_mount(const char *base_path)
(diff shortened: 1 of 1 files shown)


device/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -580,11 +686,40 @@ int32_t tud_msc_read10_cb(uint8_t lun, uint32_t lba, uint32_t offset, void *buff
...
-    esp_err_t err = msc_storage_write_sector(lba, offset, bufsize, buffer);
+#if CONFIG_IDF_TARGET_ESP32S3
+    // Ensure the buffer can accommodate the incoming data
+    if (bufsize > MSC_WRITE_BUFFER_SIZE) {
+        ESP_LOGE(TAG, "Buffer overflow: incoming data size %lu exceeds buffer size %u", bufsize, MSC_WRITE_BUFFER_SIZE);
+        return -1;
+    }
+
+    // check if previous data are still processed
+    if (write_buffer->p_write_buffer->is_data_pending) {
```
roma-jam, line 697: Seems that we don't need the pending flag here, as the task, which can trigger callback is the same task as perform writing operation. 
So there will be no new `write10` callbacks, while the data is writing to the storage.

device/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -580,11 +686,40 @@ int32_t tud_msc_read10_cb(uint8_t lun, uint32_t lba, uint32_t offset, void *buff
...
+                 write_buffer->p_write_buffer->bufsize);
+        return -1;
+    }
+    // Copy data to the buffer
+    memcpy((void *)write_buffer->p_write_buffer->data_buffer, buffer, bufsize);
+    write_buffer->p_write_buffer->lba = lba;
+    write_buffer->p_write_buffer->offset = offset;
+    write_buffer->p_write_buffer->bufsize = bufsize;
+    write_buffer->p_write_buffer->is_data_pending = true;
+    _switch_buffers();
```
roma-jam, line 711: The same thing that we don't need to swap the buffers here as it brings not that much profit.

We already have one buffer in MSC class driver, which is used by USB to place the OUT data.
The other buffer we are using as a source for the storage. 

Even if the storage will occupy the buffer to write data, USB DMA will be able to write the next OUT data to internal buffer without any collision. 

But in current implementation there will be no situations when two (three) buffers will be in use. 

So, you can remove the buffer swapping, keep only one buffer to copy data during the callback and the result will be the same: ~900 KB/s for write operation.

roma-jam (commented):
Thanks a lot for the PR! Description is perfect! Nice idea to use the resource of the same task, which runs tud_task() for writing operation. 

I pointed some general things that affects the logic. Based on the that, may I ask you to try and remove one buffer? The logic will be much simpler: get write10 callback, memcpy data, usbd_defer_func write to storage. 

As far as we don't use both buffers at the same time, this change could help us: 
- simplify the logic
- decrease memory usage (twice)
- do not loose in speed comparing to your current results

device/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -24,29 +25,69 @@
...
+/**
+ * @brief Structure managing dual write buffers and associated metadata.
+ *
+ * This structure supports a ping-pong buffering mechanism for MSC operations,
+ * allowing concurrent data processing and reception.
+ */
+typedef struct {
+    msc_write_buffer_t buffers[2]; /*!< Array of two write buffers used for ping-pong buffering. */
+    volatile msc_write_buffer_t *p_write_buffer; /*!< Pointer to the active write buffer. */
+    volatile msc_write_buffer_t *p_read_buffer;  /*!< Pointer to the active read buffer. */
```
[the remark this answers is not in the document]
igi540: Thanks for the feedback! I completely agree that using "read" terminology for a write buffer is confusing. Initially, I introduced this structure when I was experimenting with a separate RTOS task for storage writes, which is why the naming remained. Since I’ve now simplified the logic and removed the ping-pong buffering, this terminology is no longer relevant and has been cleaned up in the latest changes. 😊

device/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -580,11 +686,40 @@ int32_t tud_msc_read10_cb(uint8_t lun, uint32_t lba, uint32_t offset, void *buff
 // - Application write data from buffer to address contents (up to bufsize) and return number of written byte.
 int32_t tud_msc_write10_cb(uint8_t lun, uint32_t lba, uint32_t offset, uint8_t *buffer, uint32_t bufsize)
 {
-    esp_err_t err = msc_storage_write_sector(lba, offset, bufsize, buffer);
+#if CONFIG_IDF_TARGET_ESP32S3
+    // Ensure the buffer can accommodate the incoming data
+    if (bufsize > MSC_WRITE_BUFFER_SIZE) {
```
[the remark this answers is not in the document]
igi540: Thanks for pointing this out! You're absolutely right—since `CONFIG_TINYUSB_MSC_BUFSIZE` defines the static buffer size and the MSC class driver ensures the buffer won't exceed this limit, the size check is unnecessary.

However, instead of completely removing the check, I replaced the original `if` condition with an `assert(bufsize <= MSC_STORAGE_BUFFER_SIZE)` as a safeguard for potential future changes. If the driver behavior or buffer configuration ever changes, this assertion will help catch unexpected issues early during development. It’s mainly for defensive programming to ensure robustness without affecting runtime performance in release builds.

Thanks again for the valuable feedback! 🚀

device/esp_tinyusb/tusb_msc_storage.c
igi540: Thanks for the clarification! You're correct—since `USBD_EVENT_FUNC_CALL` ensures the same task handles both the callback and the write operation, there’s no risk of overlapping `WRITE10` callbacks while a write is still in progress.

I’ve removed the `is_data_pending` flag accordingly, which simplifies the logic without any functional impact. Appreciate your detailed feedback, it really helps streamline the code!

device/esp_tinyusb/tusb_msc_storage.c
igi540: Thanks for the suggestion! I’ve removed the buffer swapping and simplified the logic to use just a single buffer during the callback.

To confirm the impact, I’ve also rerun a few performance measurements, and the results are essentially the same as with the previous ping-pong buffer implementation—around ~900 KB/s for write operations. 🚀

Appreciate your feedback, it definitely helped clean up the code!

device/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -249,6 +276,44 @@ static esp_err_t _mount(char *drv, FATFS *fs)
     return ret;
 }
 
+static void _write_task(void *param)
```
tore-espressif, line 279: Could we rename to `_write_func` ? Since this is not a task entry function

device/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -24,29 +26,54 @@
...
+ * @brief Structure representing a single write buffer for MSC operations.
+ */
+typedef struct {
+    uint8_t data_buffer[MSC_STORAGE_BUFFER_SIZE]; /*!< Buffer to store write data. The size is defined by MSC_STORAGE_BUFFER_SIZE. */
+    uint32_t lba;                          /*!< Logical Block Address for the current WRITE10 operation. */
+    uint32_t offset;                       /*!< Offset within the specified LBA for the current write operation. */
+    uint32_t bufsize;                      /*!< Number of bytes to be written in this operation. */
+} msc_storage_buffer_t;
+
+static msc_storage_buffer_t *storage_buffer;
```
tore-espressif, line 75: Should the `storage_buffer` be part of `tinyusb_msc_storage_handle_s`? 

With current singleton implementation there is only 1 buffer, which limits future extensions to multiple MSC interfaces in 1 USB device

device/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -249,6 +276,44 @@ static esp_err_t _mount(char *drv, FATFS *fs)
...
+    if (err != ESP_OK) {
+        ESP_LOGE(TAG, "Write operation failed: 0x%x", err);
+        tud_msc_set_sense(0, SCSI_SENSE_MEDIUM_ERROR, 0x03, 0x11);// Set SCSI sense data to report a medium error to the USB host.
+    }
+}
+
+static esp_err_t _init_storage_buffer(void)
+{
+    assert(!storage_buffer);
+    storage_buffer = heap_caps_malloc(sizeof(msc_storage_buffer_t), MALLOC_CAP_DMA);
```
tore-espressif, line 297: I remember we discussed some memory alignment requirements. Aren't they needed?

tore-espressif (commented):
@igi540 Overall LGTM.

Only 1 major question about adding the storage_buffer into tinyusb_msc_storage_handle struct. This way, we would have 1 handle for 1 storage instance

roma-jam:
@igi540 and please check the CI fail: `APP: Write data after disconnect`

device/esp_tinyusb/tusb_msc_storage.c
igi540: Good point. I'll rename `_write_task` to `_write_func` to better reflect its purpose.

device/esp_tinyusb/tusb_msc_storage.c
igi540: Agreed, moving `storage_buffer` into `tinyusb_msc_storage_handle_s` makes sense for future scalability. I'll refactor the code accordingly.

device/esp_tinyusb/tusb_msc_storage.c
igi540: Yes, I’ve already addressed the memory alignment requirements in the latest implementation. I’m using `heap_caps_aligned_alloc()` with a defined alignment value (`MSC_STORAGE_MEM_ALIGN`), which is currently set to **4 bytes**. This ensures proper alignment for DMA operations.
Additionally, I’ve added a compile-time check to ensure that `CONFIG_TINYUSB_MSC_BUFSIZE` is divisible by the alignment value. I’ve also tested higher alignment values (e.g., 8, 16, and 32 bytes), but they didn’t have any noticeable impact on performance. The throughput remained consistent, so I kept the alignment at 4 bytes for efficiency and minimal memory overhead.

tore-espressif:
> @igi540 and please check the CI fail: `APP: Write data after disconnect`

This is really weird because it fails on v5.0 v5.1 and v5.2. 
It passes on v5.3, v5.4 and latest...

roma-jam:
@igi540 Not it is not related, as in this error the espressif dcd layer is used: dcd_esp32sx.c:356

We are not using it since v0.15.

tore-espressif:
@roma-jam sorry, I didn't provide any details. Very similar code with the same race condition and by-zero-division is also in the new DCD layer. However, I can't confirm it is the same problem now. Will look into it tomorrow

device/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -13,36 +13,68 @@
...
 #if SOC_SDMMC_HOST_SUPPORTED
 #include "diskio_sdmmc.h"
 #endif
 
 static const char *TAG = "tinyusb_msc_storage";
 
+#define MSC_STORAGE_MEM_ALIGN 4
+#define MSC_STORAGE_BUFFER_SIZE CONFIG_TINYUSB_MSC_BUFSIZE /*!< Size of the buffer, configured via menuconfig (MSC FIFO size) */
+#if ((MSC_STORAGE_BUFFER_SIZE) % MSC_STORAGE_MEM_ALIGN != 0)
+#error "CONFIG_TINYUSB_MSC_BUFSIZE must be divisible by MSC_STORAGE_MEM_ALIGN. Adjust your configuration (MSC FIFO size) in menuconfig."
```
peter-marcisovsky, line 31: We can also put this note (that the `CONFIG_TINYUSB_MSC_BUFSIZE` must be divisible by `MSC_STORAGE_MEM_ALIGN`)  to Kconfig help section for the `CONFIG_TINYUSB_MSC_BUFSIZE` setting.

device/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -627,6 +703,7 @@ int32_t tud_msc_scsi_cb(uint8_t lun, uint8_t const scsi_cmd[16], void *buffer, u
 // Invoked when device is unmounted
 void tud_umount_cb(void)
 {
+    usbd_cancel_deferred_func(_write_func, NULL);
```
peter-marcisovsky, line 706: Maybe a stupid question, but where is this function coming from? I could not find it in tinyusb (release/v0.17)

tore-espressif:
@igi540 The problem with race condition is fixed https://github.com/espressif/tinyusb/pull/55

Please update this PR, so we can merge it after the fix is released

roma-jam (approved):
The one thing that unleash the race condition - is immediate return from the `tud_msc_write10_cb()`. 

Maybe we could think of adding the test_app with emulated ram disk to: 
1. measure the raw speed of mscd driver speed
2. test sudden disconnection in this case (as in this case reset comes sooner than we `prepare_cbw()`, so the reset flush the `xfer->max_size`)

Otherwise, LGTM.

Please, don't forget to address all the rest notes before merge.

PR #139, opened 2025-02-21
fix(esp_tinyusb): Fix crash when external PHY is used

tore-espressif:
* Bug in esp-idf: when external phy is selected, we cannot call set_speed()

### DIFF
--- a/device/esp_tinyusb/tinyusb.c
+++ b/device/esp_tinyusb/tinyusb.c
@@ -10,7 +10,7 @@
 #include "esp_err.h"
 #include "esp_private/periph_ctrl.h"
 #include "esp_private/usb_phy.h"
-#include "soc/usb_pins.h"
+#include "soc/usb_pins.h" // TODO: Will be removed in IDF v6.0 IDF-9029
 #include "tinyusb.h"
 #include "descriptors_control.h"
 #include "tusb.h"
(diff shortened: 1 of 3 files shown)


device/esp_tinyusb/tinyusb.c
```diff
@@ -41,18 +41,36 @@ esp_err_t tinyusb_driver_install(const tinyusb_config_t *config)
...
     if (config->external_phy) {
         phy_conf.target = USB_PHY_TARGET_EXT;
         phy_conf.ext_io_conf = &ext_io_conf;
+
+        /*
+        There is a bug in esp-idf that does not allow device speed selection
+        when External PHY is used.
+        Remove this when proper fix is implemented IDF-11144
+        */
+        phy_conf.otg_speed = USB_PHY_SPEED_UNDEFINED;
```
roma-jam, line 73: I see, there is definitely one problem: 

image




We can make a quick workaround for that, adding the line: `phy_conf.otg_speed = USB_PHY_SPEED_UNDEFINED;`

and all other changes add when we will have the working example with external phy? 
Because it seems that currently, we can split it into two PRs: 
1. prevent the esp-idf crash with the external_phy=true (temp workaround, small simple, only for fixing crashes)
2. add all other important things to make it work properly (also, based on the upcoming research)

Right now it seems as a lot of comments, which might be removed  on the step 2. So maybe exclude them from step 1 and fully add on the step 2?

tore-espressif: @roma-jam Thanks, I removed changes that don't have any test coverage. PTAL!

roma-jam (approved):
LGTM!

roma-jam:
I will update the PR with the release after this is merged. Thanks.

PR #149, opened 2025-03-19
feature(usb_host_msc_test): Decrease the writing operation to sector 10

roma-jam:
## Requirements
- Function `write_read_sectors(void)` write and read only one sector with constant number 10.
- Test case `sectors_can_be_written_and_read` verifies only one sector - sector number 10. 

## Description
- Changed name of function and test to reflect the real logic of the test
- Changed the dummy operation in test case `no_background_task` to read operation only (Goal - decrease the amount of writing operations to the sector 10).

## Related
N/A

## Testing

---

## Checklist

### DIFF
--- a/host/class/msc/usb_host_msc/test_app/main/test_msc.c
+++ b/host/class/msc/usb_host_msc/test_app/main/test_msc.c
@@ -1,6 +1,6 @@
 
 /*
- * SPDX-FileCopyrightText: 2015-2024 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2025 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -281,7 +281,13 @@ static void msc_teardown(void)
     msc_test_deinit();
 }
 
-static void write_read_sectors(void)
+static void read_sector_10(void)
+{
+    uint8_t data[DISK_BLOCK_SIZE];
+    ESP_OK_ASSERT( scsi_cmd_read10(device, data, 10, 1, DISK_BLOCK_SIZE) );
+}
+
+static void write_read_verify_of_sector_10(void)
 {
     uint8_t write_data[DISK_BLOCK_SIZE];
     uint8_t read_data[DISK_BLOCK_SIZE];
@@ -319,10 +325,10 @@ TEST_CASE("sudden_disconnect", "[usb_msc]")
     msc_teardown();
 }
 
-TEST_CASE("sectors_can_be_written_and_read", "[usb_msc]")
+TEST_CASE("sector_10_can_be_written_and_read", "[usb_msc]")
 {
     msc_setup();
-    write_read_sectors();
+    write_read_verify_of_sector_10();
     msc_teardown();
 }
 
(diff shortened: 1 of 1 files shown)


host/class/msc/usb_host_msc/test_app/main/test_msc.c
```diff
@@ -530,7 +536,7 @@ TEST_CASE("no_background_task", "[usb_msc]")
     BaseType_t task_created = xTaskCreatePinnedToCore(msc_task, "msc_events", 2 * 2048, NULL, 2, NULL, 0);
     TEST_ASSERT(task_created);
     msc_test_wait_and_install_device();
-    write_read_sectors(); // Do some dummy operations
+    read_sector_10(); // Do some dummy operations
```
peter-marcisovsky, line 539: This very check (reading and writing to and from a sector and checking the read and written data) helped us to catch a recent error, which was not easiest for investigation. IMO, I would not change it.

If the only concern here is wearing out a sector, I would prefer having the original check instead having a less worn out sector. An error caused by worn out sector would be more simple to catch and we can replace a runner in no time.

roma-jam: But it still fails on setvbuf, so ...
That means we just increase the possibility to catch the problem in the test, that not suppose to catch it.
Good, but maybe it is better to make another test, which write/read to more sectors, then 1 and verify the possibility to write to all sectors, reported from the device? 

For example:  
1. Hi am a device, I have 10 sectors
2. Ok, let's try to write to every sector you have (we verify that all sectors, reported to `tud_msc_capacity_cb()` are available)
3. Write and read to all of them.

WDYT?

peter-marcisovsky: Good idea, we can extend the `write_read_sectors()`  function to actually write and read multiple sectors, as the function name says it should do.

If we had this test (checking read/write to multiple sectors), we might have caught the recent bug more easily. There might have been a better reproduce ratio.

tore-espressif: I agree with Peter that this could cripple our test coverage.

I see 2 topics here:
1. If we want to extend the test to access more sectors, we should _add_ a new test for this
2. For the existing test, implementing a simple random() function to change the accessed sector in each run could solve the problem

Just for your reference: our flash memory can withstand around 10,000 - 100,000 write cycles. So if we run the test 10x a day (we don't, though) it will fail in 3 to 30 years.

roma-jam: Thanks for the response!

Yes, Peter already pointed out that this could cripple the test coverage so I did propose to add another test. 

Also, right now the test name is `write_read_sectors(void)` could create false expectations that several sectors are under the test (which is not true).
10,000 - 100,000 write cycles - plus wear levelling and this should be no problem at all. 

Adding the randomization to the test itself is not a great approach, because it doesn't bring any value. 
In this case writing the counter to the sector will be better, because when the sector will fail on writing operation - we could always read the data from it and say how many times we did write there. 

Anyway, there was just a proposition from my side to make test better. So if you like some ideas, please let me know and I will do it. 

Otherwise we can just close this MR and keep the things they are right now.

tore-espressif: > 10,000 - 100,000 write cycles - plus wear levelling and this should be no problem at all.
Adding the randomization to the test itself is not a great approach, because it doesn't bring any value.

Correct, I forgot that we have wear_levelling underneath!

peter-marcisovsky: As pointed out, we can do here:
1. update the name of the function `write_read_sectors()` to `write_read_sector()` 
2. add a new test case which will read/write to multiple sectors more data, in case the chip support team re-enables the burst buffer on P4. So the CI will catch any possible issue with the burst buffer.

PR #150, opened 2025-03-19
test(host/uvc): Simple mock for opening UVC device

tore-espressif:
Simple UVC device opening mock

### DIFF
--- a/host/class/uvc/usb_host_uvc/host_test/main/test_fixtures.cpp
+++ b/host/class/uvc/usb_host_uvc/host_test/main/test_fixtures.cpp
@@ -1,0 +1,103 @@
+/*
+ * SPDX-FileCopyrightText: 2025 Espressif Systems (Shanghai) CO LTD
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <catch2/catch_test_macros.hpp>
+#include "mock_add_usb_device.h"
+#include "test_fixtures.hpp"
+
+extern "C" {
+#include "Mockusb_host.h"
+}
+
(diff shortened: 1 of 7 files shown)


host/class/uvc/usb_host_uvc/host_test/main/test_fixtures.cpp
```diff
@@ -0,0 +1,110 @@
...
+esp_err_t test_uvc_host_stream_open(const uvc_host_stream_config_t *stream_config, int timeout, uvc_host_stream_hdl_t *stream_hdl_ret, bool is_isoc)
+{
+    /* **Setup callbacks** */
+
+    // This will setup or mocking framework
+    usb_host_device_open_Stub(usb_host_device_open_mock_callback);
+    usb_host_get_device_descriptor_Stub(usb_host_get_device_descriptor_mock_callback);
+    usb_host_device_close_Stub(usb_host_device_close_mock_callback);
+    usb_host_get_active_config_descriptor_Stub(usb_host_get_active_config_descriptor_mock_callback);
+    usb_host_device_addr_list_fill_Stub(usb_host_device_addr_list_fill_mock_callback);
```
tore-espressif, line 70: FYI @peter-marcisovsky I found the `_Stub` CMock functions.

They work the same as `_AddCallback` but CMock stops counting the calls and stops checking the arguments.

So they seem to be suitable for setting up a mock framework, like we do.

Here I set Stubs for usb_host lib functions that I don't care about. I only need the mock_callbacks to be called

tore-espressif: This PR can be merged after the mock in esp-idf is updated

PR #165, opened 2025-04-25
refactor(esp_tinyusb): Descriptors configuration

roma-jam:
## Description

Refactor tinyUSB configuration with: 
- Split the configuration parameters by 4 groups: port-related, task-related, descriptors-related and phy-related. 
- Renamed descriptor-configuration structure to descriptors map and move descriptor configuration to the task. (Fixes the problem with breaking the descriptors, while another task is called the `tinyusb_driver_install`)

## Testing

- Enabled test `Multitask install`, which verifies the possibility to install tinyusb driver from five tasks simultaneously. 

---

## Checklist


device/esp_tinyusb/include/tinyusb.h
```diff
@@ -20,41 +20,34 @@ extern "C" {
  * @brief Configuration structure of the TinyUSB core
  */
 typedef struct {
-    // USB Peripheral Settings
-    bool skip_phy_setup;                       /*!< If set, the esp_tinyusb will not configure the USB PHY thus allowing
+    tinyusb_port_t port;                        /*!< USB Peripheral hardware port number. Available when hardware has several available peripherals. */
+    struct {
+        // USB Peripheral Settings
+        bool skip_setup;                       /*!< If set, the esp_tinyusb will not configure the USB PHY thus allowing
```
igi540, line 26: Would it make sense to keep the original name `skip_phy_setup` here for consistency with the previous versions and to make its purpose a bit clearer?

roma-jam: This is within a breaking change anyway.

When we choose between: 
1. `config->phy.skip_setup = true; `
2. `config->phy.skip_phy_setup = true;`

seems that the usage of two `phy`s in the name is just not necessary.

device/esp_tinyusb/include/tinyusb.h
```diff
@@ -20,41 +20,34 @@ extern "C" {
...
-        };
-    };
-    const uint8_t *hs_configuration_descriptor;                 /*!< Pointer to a HighSpeed configuration descriptor. If set to NULL, TinyUSB device will use a default configuration descriptor whose values are set in Kconfig */
-    const tusb_desc_device_qualifier_t *qualifier_descriptor;   /*!< Pointer to a qualifier descriptor */
-#else
-    };
-#endif // TUD_OPT_HIGH_SPEED
+    struct {
+        const tusb_desc_device_t *device;           /*!< Pointer to a device descriptor. If set to NULL, the TinyUSB device will use a default device descriptor whose values are set in Kconfig */
+        const tusb_desc_device_qualifier_t *qualifier;   /*!< Pointer to a qualifier descriptor */
```
peter-marcisovsky, line 44: In the USB Specification, there is always used term `device qualifier descriptor` and never used term just `qualifier descriptor`. Do you think it would make sense to have `device_qualifier` instead of just `qualifier` ?

`config->descriptor->qualifier`
`config->descriptor->device_qualifier`

Even though, there is no other descriptor which holds `qualifier` in it's name, so using just qualifier here is explanatory enough. But having the `device_qualifier` would be more for the sake of completeness maybe.

device/esp_tinyusb/include/tinyusb.h
igi540: Thanks, that makes sense. I agree, it’s clear enough inside `phy`.

device/esp_tinyusb/descriptors_control.c
```diff
@@ -160,78 +148,102 @@ uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
...
+        if (config->high_speed_config == NULL) {
+#if (CFG_TUD_CDC > 0 || CFG_TUD_MSC > 0 || CFG_TUD_NCM > 0)
+            // We provide default config descriptors only for CDC, MSC and NCM classes
+            ESP_LOGW(TAG, "No High-speed configuration descriptor provided, using default.");
+            s_desc_cfg.hs_cfg = descriptor_hs_cfg_default;
 #else
-        ESP_LOGW(TAG, "No HighSpeed configuration descriptor provided, using default.");
-        s_desc_cfg.hs_cfg = descriptor_hs_cfg_default;
+            // Default configuration descriptor must be provided for the following classes
+            ESP_GOTO_ON_FALSE(config->high_speed_config, ESP_ERR_INVALID_ARG, fail, TAG, "Full-speed configuration descriptor must be provided for this device");
```
peter-marcisovsky, line 211:
```suggestion
            ESP_GOTO_ON_FALSE(config->high_speed_config, ESP_ERR_INVALID_ARG, fail, TAG, "High-speed configuration descriptor must be provided for this device");
```

device/esp_tinyusb/tinyusb.c
```diff
@@ -43,40 +43,39 @@ static esp_err_t tinyusb_check_config(const tinyusb_config_t *config)
...
+    if (!config->phy.skip_setup) {
         // Configure USB PHY
         usb_phy_config_t phy_conf = {
             .controller = USB_PHY_CTRL_OTG,
             .target = USB_PHY_TARGET_INT,
             .otg_mode = USB_OTG_MODE_DEVICE,
             // OTG speed selects the PHY indirectly
+#if ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 3, 0)
+            // esp32p4 is supported from IDF 5.3
             .otg_speed = (config->port == TINYUSB_PORT_0) ? USB_PHY_SPEED_FULL : USB_PHY_SPEED_HIGH,
```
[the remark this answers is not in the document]
roma-jam: Not really. 
The only reason why do we need to provide the speed is to make the configuration `usb_phy_otg_dev_set_speed()` and do the pull override. But we do not need to do the pull override. 

The default value is `USB_PHY_SPEED_UNDEFINED` and with this value we do no do the override, so we need to setup the speed only for selecting the right phy...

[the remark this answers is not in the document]
roma-jam: ... but thanks for pointing attention to that. 
I revisit it again and rewrite the way how we configure the phy, based on the port number. 
As we need to use the default PHY for every peripheral, so it is like this right now: 
- P4, + HS Periph = UTMI PHY
- P4 + FS Periph = INT PHY
- S2/S3 + FS Periph = INT PHY

device/esp_tinyusb/tinyusb_task.c
```diff
@@ -82,24 +108,32 @@ esp_err_t tinyusb_task_check_config(const tinyusb_task_config_t *config)
...
 
     esp_err_t ret;
-    // Allocate TinyUSB Task object
     tinyusb_task_ctx_t *task_ctx = heap_caps_calloc(1, sizeof(tinyusb_task_ctx_t), MALLOC_CAP_DEFAULT);
     if (task_ctx == NULL) {
         return ESP_ERR_NO_MEM;
     }
 
     task_ctx->awaiting_handle = xTaskGetCurrentTaskHandle();    // Save parent task handle
+    task_ctx->p_mux_lock = &s_mux_lock;                         // Save mutex lock
```
tore-espressif, line 131: This looks dangerous. Do we use the same semaphore to guard 2 different things?

device/esp_tinyusb/tinyusb_task.c
```diff
@@ -82,24 +108,32 @@ esp_err_t tinyusb_task_check_config(const tinyusb_task_config_t *config)
...
 {
     TINYUSB_TASK_ENTER_CRITICAL();
     TINYUSB_TASK_CHECK_FROM_CRIT(p_tusb_task_ctx == NULL, ESP_ERR_INVALID_STATE);             // Task shouldn't started
+    if (s_mux_lock == NULL) {
+        s_mux_lock = xSemaphoreCreateMutex();
+        if (s_mux_lock == NULL) {
+            TINYUSB_TASK_EXIT_CRITICAL();
+            return ESP_ERR_NO_MEM;
+        }
+    }
```
tore-espressif, line 121: In case `s_mux_lock != NULL` we will continue with TinyUSB initialization. Is that intended?

device/esp_tinyusb/include/tinyusb_task.h
```diff
@@ -51,7 +52,7 @@ esp_err_t tinyusb_task_check_config(const tinyusb_task_config_t *task_cfg);
  *    - ESP_ERR_NO_MEM if memory allocation failed
  *    - ESP_OK if TinyUSB Task initialized successfully
  */
-esp_err_t tinyusb_task_start(tinyusb_port_t port, const tinyusb_task_config_t *task_cfg);
+esp_err_t tinyusb_task_start(tinyusb_port_t port, const tinyusb_task_config_t *task_cfg, const tinyusb_desc_config_t *desc_cfg);
```
tore-espressif, line 55: Should this API be private now?

tore-espressif (commented):
@roma-jam Thank you for the update!
I left few comments so you're not blocked

device/esp_tinyusb/descriptors_control.c
```diff
@@ -160,78 +148,102 @@ uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
...
-    // Qualifier Descriptor
-    if (config->qualifier_descriptor == NULL) {
-        ESP_GOTO_ON_FALSE((s_desc_cfg.dev == &descriptor_dev_default), ESP_ERR_INVALID_ARG, fail, TAG, "Qualifier descriptor must be present (Device Descriptor not default).");
-        // Get default qualifier if device descriptor is default
-        ESP_LOGW(TAG, "No Qulifier descriptor provided, using default.");
-        s_desc_cfg.qualifier = &descriptor_qualifier_default;
+        // High-speed and Full-speed descriptors should be equal length
+        ESP_GOTO_ON_FALSE(((tusb_desc_configuration_t *)s_desc_cfg.hs_cfg)->wTotalLength ==
+                          ((tusb_desc_configuration_t *)s_desc_cfg.fs_cfg)->wTotalLength,
+                          ESP_ERR_INVALID_ARG, fail, TAG, "High-speed and Full-speed configuration descriptors must be same length");
```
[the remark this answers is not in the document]
roma-jam: Yep, they might be completely different, right. 
Thanks.

device/esp_tinyusb/descriptors_control.c
```diff
@@ -160,78 +148,102 @@ uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
...
+        // Device Qualifier Descriptor
+        if (config->qualifier == NULL) {
+            // Get default qualifier if device descriptor is default
+            ESP_LOGW(TAG, "No Qualifier descriptor provided, using default.");
+            s_desc_cfg.qualifier = &descriptor_qualifier_default;
+        } else {
+            s_desc_cfg.qualifier = config->qualifier;
+        }
+        // Other Speed Descriptor buffer allocation, will be used for other speed configuration descriptor request
+        s_desc_cfg.other_speed = calloc(1, ((tusb_desc_configuration_t *)s_desc_cfg.hs_cfg)->wTotalLength);
```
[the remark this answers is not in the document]
roma-jam: In case, fs and hs config descriptors could have different sizes, we need to allocate the memory for the bigger one.

device/esp_tinyusb/include/tinyusb_task.h
roma-jam: Yes, indeed. 
There is no point to keep it public. Thanks, changed.

device/esp_tinyusb/descriptors_control.c
```diff
@@ -160,78 +148,102 @@ uint16_t const *tud_descriptor_string_cb(uint8_t index, uint16_t langid)
...
+    if (port == TINYUSB_PORT_1) {
+#if (TUD_OPT_HIGH_SPEED)
+        //
+#else
+        ESP_RETURN_ON_FALSE(port == TINYUSB_PORT_0, ESP_ERR_INVALID_ARG, TAG, "Only one port is available");
+#endif // TUD_OPT_HIGH_SPEED
+    }
+#endif // SOC_USB_OTG_PERIPH_NUM > 1
+    return ESP_OK;
+}
```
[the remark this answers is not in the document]
roma-jam: Yes, I changed it to: 

```
#if (SOC_USB_OTG_PERIPH_NUM > 1)
    if (port == TINYUSB_PORT_1) {
#if !TUD_OPT_HIGH_SPEED
        ESP_RETURN_ON_FALSE(false, ESP_ERR_INVALID_ARG, TAG, "Device has only Full-speed port");
#endif
    }
#endif
```

The `TINYUSB_PORT_1` isn't available for user when `SOC_USB_OTG_PERIPH_NUM == 1`, but that still possible if someone will use just `1`

device/esp_tinyusb/tinyusb_task.c
roma-jam: I removed the mux completely, simplified the logic. 
Now, we have: 

`static bool _task_is_running = false;               // Locking flag for the task, access only from the critical section`

which plays the same role as the mutex before, but way more simple.

device/esp_tinyusb/tinyusb_task.c
roma-jam: No, not really. 
We handle this case later...

I simplified the logic and now we lock the task till the moment it will start or fail. 
While the task is locked, we don't let the user to start more tasks.

device/esp_tinyusb/tinyusb_task.c
```diff
@@ -38,35 +39,61 @@ typedef struct {
...
+        goto del;
+    }
     if (!tusb_rhport_init(task_ctx->rhport, &task_ctx->rhport_init)) {
         ESP_LOGE(TAG, "Init TinyUSB stack failed");
-        vTaskDelete(NULL);
-        // No return needed here: vTaskDelete(NULL) does not return
+        goto desc_free;
     }
+
     ESP_LOGD(TAG, "TinyUSB stack has been initialized in task");
```
tore-espressif, line 77: Since initializing TinyUSB in task is now the only option, could we remove this log?

device/esp_tinyusb/tinyusb_task.c
tore-espressif: Looks good!

tore-espressif (commented):
@roma-jam thanks for the update, looks better now!

It seems that the TinyUSB port <-> USB peripheral is still not clear to me... Could you also please sum it up in a README table for users?

device/esp_tinyusb/include_private/tinyusb_task.h
```diff
@@ -51,7 +43,7 @@ esp_err_t tinyusb_task_check_config(const tinyusb_task_config_t *task_cfg);
  *    - ESP_ERR_NO_MEM if memory allocation failed
  *    - ESP_OK if TinyUSB Task initialized successfully
  */
-esp_err_t tinyusb_task_start(tinyusb_port_t port, const tinyusb_task_config_t *task_cfg);
+esp_err_t tinyusb_task_start(tinyusb_port_t port, const tinyusb_task_config_t *task_cfg, const tinyusb_desc_config_t *desc_cfg);
```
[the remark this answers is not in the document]
roma-jam: Before (when there was a possibility to init the task externally), descriptors and task could be possible to have at the same logic level. 
Now, when there is only one option possible (run the task internally), there is no point to configure descriptors outside of the task, so the level of descriptors could be increased and it might be used only from the task. 

So, before it was: 
```
	 +--------------+
	 |  application |
	 +--------------+
	        | 
	 +--------------+
	 |  tinyusb.c   |
	 +--------------+
	    |       |
      +-----+       +---+
      |                 |
+----------+   +----------------+  
|  task.c  |   |  descriptors.c |
+----------+   +----------------+ 
```

Right now, there is no need to configure/re-configure the descriptors, if the task is already running. Moreover, this shouldn't be possible to do, as it should be locked till the moment the task is running. 
So, it can be represent as following: 

```
	 +----------------+
	 |  application.c |
	 +----------------+
	        | 
	 +--------------+
	 |  tinyusb.c   |
	 +--------------+
	        |         
	  +------------+   
	  |  task.c    |  
	  +------------+  
                | 
  	 +----------------+ 
 	 |  descriptors.c |
 	 +----------------+ 
```
 

So, right now it is still Single Responsibility Principle, except maybe the `tinyusb_descriptors_check()` which I moved to the `tinyusb_driver_install()` just to check the descriptors in more explicit way.
But sure, I could move it to the task.c, so there will be no exceptions and the principle will be hold unquestionable.

device/esp_tinyusb/tinyusb.c
```diff
@@ -43,40 +43,46 @@ static esp_err_t tinyusb_check_config(const tinyusb_config_t *config)
...
+#if (SOC_USB_OTG_PERIPH_NUM > 1)
+        if (config->port == TINYUSB_PORT_1) {
+            // Default PHY for OTG2.0 is UTMI
+            phy_conf.target = USB_PHY_TARGET_UTMI;
+            phy_conf.otg_speed = USB_PHY_SPEED_HIGH;
+        } else if (config->port == TINYUSB_PORT_0) {
+            // Default PHY for OTG1.1 is INT
+            phy_conf.otg_speed = USB_PHY_SPEED_FULL;
+        }
+#endif // (SOC_USB_OTG_PERIPH_NUM > 1)
```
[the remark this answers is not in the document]
roma-jam: The mapping in the tinyusb is the opposite, regarding the mapping we have in esp-idf (https://github.com/hathach/tinyusb/blob/master/src/portable/synopsys/dwc2/dwc2_esp32.h#L56): 

```
// On ESP32 for consistency we associate
// - Port0 to OTG_FS, and Port1 to OTG_HS
static const dwc2_controller_t _dwc2_controller[] = {
  { .reg_base = DWC2_FS_REG_BASE, .irqnum = ETS_USB_OTG11_CH0_INTR_SOURCE, .ep_count = 7, .ep_in_count = 5, .ep_fifo_size = 1024 },
  { .reg_base = DWC2_HS_REG_BASE, .irqnum = ETS_USB_OTG_INTR_SOURCE, .ep_count = 16, .ep_in_count = 8, .ep_fifo_size = 4096 }
};
```

Which means, that for tinyusb port 0 - is FS peripheral and port 1 - is HS. 


So, if we want the same numbers for the peripherals as in esp-idf, I need to invert them somewhere. 
Right now, I am not inverting them, just not to add additional difficulties.

I see two options right now: 
1. Remove the number from the tinyusb port enum. So to say, to change `TINYUSB_PORT_0` -> `TINYUSB_PORT_FS`. This will exclude the using of numbers and confusion. But there could be situation, that some hardware in the future could have two HS periph and in this case we will need to change it again. 
2. Invert the `TINYUSB_PORT_*`  so it will be the same, as in esp-idf: `*_0` - for HS, `*_1` - for FS. In this case, it will be the same, we can add the any periph in the future without problems, even if there will be two HS periphs. 

I would prefer the first option, because it is much clear to the user which peripheral he uses (FS or HS). So I need one more vote, because by the arguments seems that the second option is more preferable. 

Feel free to vote, but apparently I will change it according to the option #2.

device/esp_tinyusb/include_private/tinyusb_task.h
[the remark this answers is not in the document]
roma-jam: > Now task and descriptors are mixed together

I assume, that the header-file `tinyusb_types.h` where we have configuration structs for task and descriptors create this image, that they are mixed. I removed that file and move the configuration to the `tinyusb.h` file, which is: 
- public
- should be used when nothing special required

Also, I moved the descriptor verification to the task logic. 
 
I hope that will help to keep the idea more transparent.

device/esp_tinyusb/tinyusb.c
```diff
@@ -43,40 +43,46 @@ static esp_err_t tinyusb_check_config(const tinyusb_config_t *config)
...
+    usb_phy_handle_t phy_hdl = NULL;
+    if (!config->phy.skip_setup) {
         // Configure USB PHY
         usb_phy_config_t phy_conf = {
             .controller = USB_PHY_CTRL_OTG,
             .target = USB_PHY_TARGET_INT,
             .otg_mode = USB_OTG_MODE_DEVICE,
-            // OTG speed selects the PHY indirectly
-            .otg_speed = (config->port == TINYUSB_PORT_0) ? USB_PHY_SPEED_FULL : USB_PHY_SPEED_HIGH,
+            .otg_speed = USB_PHY_SPEED_UNDEFINED,
```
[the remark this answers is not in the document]
roma-jam: Yes, I started with it, but if I config the parameters correctly, based on the peripheral - it is not correct. 
Changed it to: 

```
        // Configure USB PHY
        usb_phy_config_t phy_conf = {
            .controller = USB_PHY_CTRL_OTG,
            .target = USB_PHY_TARGET_INT,
            .otg_mode = USB_OTG_MODE_DEVICE,
            .otg_speed = USB_PHY_SPEED_FULL,
        };

#if (SOC_USB_OTG_PERIPH_NUM > 1)
        if (config->port == TINYUSB_PORT_HIGH_SPEED_0) {
            // Default PHY for OTG2.0 is UTMI
            phy_conf.target = USB_PHY_TARGET_UTMI;
            phy_conf.otg_speed = USB_PHY_SPEED_HIGH;
        }
#endif // (SOC_USB_OTG_PERIPH_NUM > 1)
```

device/esp_tinyusb/tinyusb.c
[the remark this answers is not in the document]
roma-jam: I took the option 1, and added there a number as an option to extend it in the future: 
```
typedef enum {
    TINYUSB_PORT_FULL_SPEED_0 = 0,                /*!< USB OTG 1.1 peripheral */
#if (SOC_USB_OTG_PERIPH_NUM > 1)
    TINYUSB_PORT_HIGH_SPEED_0,                    /*!< USB OTG 2.0 peripheral */
#endif // (SOC_USB_OTG_PERIPH_NUM > 1)
    TINYUSB_PORT_MAX,
} tinyusb_port_t;
```

So, the user would operate with `TINYUSB_PORT_FULL_SPEED_0` and `TINYUSB_PORT_HIGH_SPEED_0` (or `TINYUSB_PORT_HIGH_SPEED_1`) which is easier to understand and the name does not interfere with just a number (0 and 1 in esp-idf) or a port number (0 and 1 in TinyUSB).

igi540:
I've tested the changes from this PR on both **ESP32-S3** and **ESP32-P4** hardware, using the updated **test_apps/cdc/main/test_cdc.c** example included in the PR. Everything works as expected — the TinyUSB descriptors have been refactored cleanly, the device enumerates and communicates correctly, and the configuration changes are functional.

This refactor improves code readability, structure, and multi-port (FS/HS) support for ESP32 series. Thanks for the work on this.

igi540 (approved):
LGTM.

PR #122, opened 2025-01-20
feat(tusb_msc): Add device mount/unmount callback

peter-marcisovsky:
## Description

This MR adds a possibility to define a user callback for device mount/unmount events.

Right now, the `tusb_msc_storage.c` automatically calls:
- `tinyusb_msc_storage_mount()` at `tud_umount_cb()`
- `tinyusb_msc_storage_unmount()` at `tud_mount_cb()`

User has no flexibility to do anything else upon the device mount/unmount.

This MR adds a flexibility to the user, to for example control an owership of the storage (eg, not to automatically unmount a storage when a USB device is plugged in)

Also cleaned conceptions between device mount/unmount and storage mount/unmount, not to confuse anyone when the code refers to device mount/unmount and when to storage mount/unmount.

## Related

Related to #78 

## Testing

---

## Checklist

Before submitting a Pull Request, please ensure the following:

- [ ] 🚨 This PR does not introduce breaking changes.
- [ ] All CI checks (GH Actions) pass.
- [ ] Documentation is updated as needed.
- [ ] Tests are updated or added as necessary.
- [ ] Code is well-commented, especially in complex areas.
- [ ] Git history is clean — commits are squashed to the minimum necessary.

PR #175, opened 2025-05-22
feature(tinyusb): Added device events callback

roma-jam:
## Description

User could specify the specific callback, that will be called with specific event:
- `TINYUSB_EVENT_ATTACHED`: Host enumerates the device and fulfill SetConfiguration() request on a device. 
- `TINYUSB_EVENT_DETACHED`: Device detaches from the Host. 

In the future, events list could be extended to: `SUSPEND`, `RESUME`, `SOF` etc.

## Related
- Prerequisites for https://github.com/espressif/esp-usb/pull/178
- https://github.com/espressif/esp-usb/pull/122

## Testing

Testing is possible with the same tests.

---

## Checklist

### DIFF
--- a/device/esp_tinyusb/include/tinyusb_default_configs.h
+++ b/device/esp_tinyusb/include/tinyusb_default_configs.h
@@ -14,6 +14,8 @@
 extern "C" {
 #endif
 
+#define GET_CONFIG_MACRO(dummy, arg1, arg2, arg3, name, ...)    name
+
 /**
  * @brief Default TinyUSB Driver configuration structure initializer
  *
@@ -31,10 +33,24 @@ extern "C" {
  * - Unicore:       CPU0
  *
  */
+
+#define TINYUSB_DEFAULT_CONFIG(...)              GET_CONFIG_MACRO(, ##__VA_ARGS__, \
+                                                                    TINYUSB_CONFIG_INVALID,    \
+                                                                    TINYUSB_CONFIG_EVENT_ARG,  \
+                                                                    TINYUSB_CONFIG_EVENT,      \
+                                                                    TINYUSB_CONFIG_NO_ARG      \
+                                                                )(__VA_ARGS__)
+
+#define TINYUSB_CONFIG_INVALID(...)              static_assert(false, "Too many arguments for TINYUSB_DEFAULT_CONFIG")
+
 #if CONFIG_IDF_TARGET_ESP32P4
-#define TINYUSB_DEFAULT_CONFIG()       TINYUSB_CONFIG_HIGH_SPEED()
+#define TINYUSB_CONFIG_NO_ARG()                  TINYUSB_CONFIG_HIGH_SPEED(NULL, NULL)
+#define TINYUSB_CONFIG_EVENT(event_hdl)          TINYUSB_CONFIG_HIGH_SPEED(event_hdl, NULL)
+#define TINYUSB_CONFIG_EVENT_ARG(event_hdl, arg) TINYUSB_CONFIG_HIGH_SPEED(event_hdl, arg)
 #else
-#define TINYUSB_DEFAULT_CONFIG()       TINYUSB_CONFIG_FULL_SPEED()
+#define TINYUSB_CONFIG_NO_ARG()                  TINYUSB_CONFIG_FULL_SPEED(NULL, NULL)
+#define TINYUSB_CONFIG_EVENT(event_hdl)          TINYUSB_CONFIG_FULL_SPEED(event_hdl, NULL)
+#define TINYUSB_CONFIG_EVENT_ARG(event_hdl, arg) TINYUSB_CONFIG_FULL_SPEED(event_hdl, arg)
 #endif
 
 #if CONFIG_FREERTOS_UNICORE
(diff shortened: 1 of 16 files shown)


device/esp_tinyusb/include/tinyusb_default_configs.h
```diff
@@ -48,9 +64,7 @@ extern "C" {
 // Default priority for task used in TinyUSB task creation
 #define TINYUSB_DEFAULT_TASK_PRIO      5
 
-
-
-#define TINYUSB_CONFIG_FULL_SPEED()                     \
+#define TINYUSB_CONFIG_FULL_SPEED(event_hdl, arg)       \
```
tore-espressif, line 67: I'm not sure if this change is necessary. Many users will not use the new `event_cb` so they will end up with calling `TINYUSB_CONFIG_FULL_SPEED(NULL, NULL)`

tore-espressif (commented):
@roma-jam Only few details, overall LGTM

device/esp_tinyusb/include/tinyusb_default_configs.h
roma-jam: They still could call `TINYUSB_DEFAULT_CONFIG()` without any arguments and still have the same result. 

I didn't apply similar logic for `TINYUSB_CONFIG_FULL_SPEED()` because I assumed that calling this API will require some knowledge, so generally it is not a big deal to provide `NULL` as argument.

tore-espressif: Fair enough, thanks!

device/esp_tinyusb/include/tinyusb_types.h
```diff
@@ -26,6 +26,23 @@ typedef enum {
...
+/**
+ * @brief TinyUSB event type
+ */
+typedef enum {
+    TINYUSB_EVENT_ATTACHED,             /*!< USB device attached to the Host */
+    TINYUSB_EVENT_DETACHED,             /*!< USB device detached from the Host */
+} tinyusb_event_t;
+
+/**
+ * @brief Callback used to indicate that the External Port driver requires a Hub class specific request
```
peter-marcisovsky, line 50: Probably a copy-paste error from an external hub callback

device/esp_tinyusb/tinyusb.c
```diff
@@ -14,17 +14,50 @@
...
 #ifndef tusb_teardown
 #   define tusb_teardown()   (true)
 #endif // tusb_teardown
 
+// Application callbacks
+
+void tud_mount_cb(void)
+{
+#if (CONFIG_TINYUSB_MSC_ENABLED)
+    tinyusb_msc_mount();
```
peter-marcisovsky, line 47: Is this going to be the storage mount function? 
Equivalent, to what we have now `tinyusb_msc_storage_mount()` in `tusb_msc_storage.c` ?

If so, it's again inverted. It should be: 

Device mount event -> Storage unmount (Unmount from an app)
Device mount event -> Storage mount (Mount to an app)

peter-marcisovsky (commented):
Nice changes.

device/esp_tinyusb/tinyusb.c
roma-jam: Nope, this is right this time.

Indeed, it should be that same as before, without any changes: 

```
// Invoked when device is unmounted
void tud_umount_cb(void)
{
    if (tinyusb_msc_storage_mount(s_storage_handle->base_path) != ESP_OK) {
        ESP_LOGW(TAG, "tud_umount_cb() mount Fails");
    }
}

// Invoked when device is mounted (configured)
void tud_mount_cb(void)
{
    tinyusb_msc_storage_unmount();
}
```

So, as the `s_storage_handle` isn't available I have already added two additional calls in `tusb_msc_storage.h`, which will be solving the logic in `tusb_msc_storage.c` correctly: 

```
// Invoked when device is unmounted
void tinyusb_msc_unmount(void)
{
    if (tinyusb_msc_storage_mount(s_storage_handle->base_path) != ESP_OK) {
        ESP_LOGW(TAG, "tud_umount_cb() mount Fails");
    }
}

// Invoked when device is mounted (configured)
void tinyusb_msc_mount(void)
{
    tinyusb_msc_storage_unmount();
} 

```

But from the tinyusb.c point of view, the logic is linear and maybe to avoid such mess in the future, we need to rename them to something like: 

`tud_mount_cb() `-> `tinyusb_msc_storage_to_usb()`
`tud_umount_cb()` -> `tinyusb_msc_storage_to_app()`

device/esp_tinyusb/include/tinyusb_types.h
roma-jam: Ah, you are totally right )

device/esp_tinyusb/tinyusb.c
roma-jam: Yep, I implemented (https://github.com/espressif/esp-usb/pull/178) two functions: 
- `void tinyusb_msc_storage_to_app(void)`
- `void tinyusb_msc_storage_to_usb(void)`

So, there should be no questions anymore.

PR #176, opened 2025-05-22
UVC updates for esp_video

tore-espressif:
## UVC driver updates for esp_video binding
Espressif's [esp_video](https://components.espressif.com/components/espressif/esp_video) component provides video pipeline and simulates V4L2 API. This MR updates UVC driver so it can be used in UVC video

### Notable changes
* Added `uvc_host_stream_format_get()` and `uvc_host_buf_info_get()` functions that are needed for esp_video_uvc binding
* Added option to request default FPS and/or default format during format change
* Fixed minor bugs

### DIFF
--- a/host/class/uvc/usb_host_uvc/host_test/main/parsing/test_parsing_helpers.hpp
+++ b/host/class/uvc/usb_host_uvc/host_test/main/parsing/test_parsing_helpers.hpp
@@ -7,7 +7,7 @@
 #pragma once
 
 /**
- * @brief Define MJPEG formats with FPS 30-5
+ * @brief Define MJPEG formats with FPS 30-5 and 0 (default)
  */
 #define FORMAT_MJPEG_30_5(x, y)      \
     {x, y, 30, UVC_VS_FORMAT_MJPEG}, \
@@ -19,7 +19,7 @@
     {x, y, 0, UVC_VS_FORMAT_MJPEG}
 
 /**
- * @brief Define MJPEG formats with FPS 30-15
+ * @brief Define MJPEG formats with FPS 30-15 and 0 (default)
  */
 #define FORMAT_MJPEG_30_25_15(x, y)  \
     {x, y, 30, UVC_VS_FORMAT_MJPEG}, \
@@ -28,7 +28,7 @@
     {x, y, 0, UVC_VS_FORMAT_MJPEG}
 
 /**
- * @brief Define MJPEG formats with FPS 30-15
+ * @brief Define MJPEG formats with FPS 30-15 and 0 (default)
  */
 #define FORMAT_MJPEG_30_20_15(x, y)  \
     {x, y, 30, UVC_VS_FORMAT_MJPEG}, \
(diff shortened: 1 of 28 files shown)


host/class/uvc/usb_host_uvc/uvc_control.c
```diff
@@ -62,7 +62,24 @@ static esp_err_t uvc_host_stream_control(uvc_host_stream_hdl_t stream_hdl, uvc_v
...
+            case UVC_VS_DESC_SUBTYPE_FRAME_FRAME_BASED:
+                vs_control->dwFrameInterval = frame_desc->frame_based.dwDefaultFrameInterval;
+                break;
+            case UVC_VS_DESC_SUBTYPE_FRAME_UNCOMPRESSED:
+            case UVC_VS_DESC_SUBTYPE_FRAME_MJPEG:
+                vs_control->dwFrameInterval = frame_desc->mjpeg_uncompressed.dwDefaultFrameInterval;
+                break;
+            default:
+                assert(!"This subtype is not supported!");
+                return ESP_ERR_NOT_SUPPORTED;
```
peter-marcisovsky, line 78: Just out of curiosity: Is this common practice, to call `assert(false)` and afterwards a `return` ?

Also, for debug builds, we have the assert and message, whereas for release builds, the asserts are removed and the code returns `ESP_ERR_NOT_SUPPORTED` but the return value of `uvc_host_stream_control()` is not always checked.

For example, `uvc_host_stream_control_probe()` (which is also called from UVC Host public API) calls `uvc_control_probe_set()` which calls `uvc_host_stream_control()`, but the return value of the `uvc_control_probe_set()` is not checked. So the code can fail without any sign of failing.

peter-marcisovsky (approved):
Just few comments for tests. LGTM

host/class/uvc/usb_host_uvc/host_test/main/parsing/test_parsing_helpers.hpp
```diff
@@ -15,23 +15,26 @@
     {x, y, 20, UVC_VS_FORMAT_MJPEG}, \
     {x, y, 15, UVC_VS_FORMAT_MJPEG}, \
     {x, y, 10, UVC_VS_FORMAT_MJPEG}, \
-    {x, y, 5, UVC_VS_FORMAT_MJPEG}
+    {x, y, 5, UVC_VS_FORMAT_MJPEG},  \
+    {x, y, 0, UVC_VS_FORMAT_MJPEG}
```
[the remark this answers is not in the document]
tore-espressif: OK, I will update the macro name and description.

I have defined macros only for the most common formats. The remaining formats are explicitely listed in the test itself

host/class/uvc/usb_host_uvc/uvc_control.c
tore-espressif: You're right. In this case, asserting by default is not the right solution.

`assert` in default switch is OK for defensive programming, where the default value cannot be ever reached.

Here however, the default switch is reachable if user plugs in a camera with format that is not supported by this driver -> will fix.

Later in `uvc_host_stream_control_probe()` I do not check the return code, because the result of the `set` operation is checked by `uvc_control_probe_get()` which checks if the camera really accepted the set format.

PR #177, opened 2025-05-23
fix(uvc): Relax bulk EoF flag check

tore-espressif:
Closes IDF-12702

Alse clean-up the streaming host tests


tore-espressif:
cc @lijunru-hub

PR #178, opened 2025-05-23
feat(tusb_msc): Add device mount/unmount callback

roma-jam:
## Testing

---

## Checklist

Before submitting a Pull Request, please ensure the following:

- [ ] 🚨 This PR does not introduce breaking changes.
- [ ] All CI checks (GH Actions) pass.
- [ ] Documentation is updated as needed.
- [ ] Tests are updated or added as necessary.
- [ ] Code is well-commented, especially in complex areas.
- [ ] Git history is clean — commits are squashed to the minimum necessary.

PR #186, opened 2025-06-13
fix(tinyusb_msc): Address calculation verification for SPI Flash

roma-jam:
## Description
The verification of the address value (the value, which is used for Wear Levelling above SPI Flash storage) is done when WRITE(10) command is coming, for both storages: SPI Flash and SD/MMC.

This verification is not relevant, when SD/MMC storage is used, as SD/MMC storage works with LBA directly. 

## Related

- Closes https://github.com/espressif/esp-idf/issues/16085
- Relates to https://github.com/espressif/esp-usb/pull/178

## Testing

```
$ dd if=/dev/random of=/dev/sda bs=512 count=1 seek=124735487
$ dd if=/dev/sda iflag=direct bs=512 skip=124735487 count=1
```

1. After the change, the direct read/write of the last sector of the card with capacity of 64 GB completed successful and the data is valid. 
image

2. Without change, the direct read/write of the last sector of the uSD card with capacity of 64 GB completed successful, but the data has not been written correctly. 
image

Notes
1. The tests were done one by one, so the returning data in the case 2 (without changes of this PR) is equal to the random data, written in case 1. 
2. Different uSD cards with 64 GB could report different amounts of sectors. In case of tested uSD card, the card capacity was  `capacity=124735488`

## Limitation

1. Reporting that data has been written correctly, even if the data was not written. Will be fixed in https://github.com/espressif/esp-usb/pull/178 

---

## Checklist

### DIFF
--- a/device/esp_tinyusb/tusb_msc_storage.c
+++ b/device/esp_tinyusb/tusb_msc_storage.c
@@ -137,9 +137,13 @@ static esp_err_t _write_sector_spiflash(size_t sector_size,
                                         size_t size,
                                         const void *src)
 {
-    ESP_RETURN_ON_ERROR(wl_erase_range(s_storage_handle->wl_handle, addr, size),
-                        TAG, "Failed to erase");
-    return wl_write(s_storage_handle->wl_handle, addr, src, size);
+    (void) addr; // addr argument is not used in this function, we calculate it based on lba and offset.
+    size_t temp = 0;
+    size_t src_addr = 0; // Address of the data to be write, relative to the beginning of the partition.
+    ESP_RETURN_ON_FALSE(!__builtin_umul_overflow(lba, sector_size, &temp), ESP_ERR_INVALID_SIZE, TAG, "overflow lba %lu sector_size %u", lba, sector_size);
+    ESP_RETURN_ON_FALSE(!__builtin_uadd_overflow(temp, offset, &src_addr), ESP_ERR_INVALID_SIZE, TAG, "overflow addr %u offset %lu", temp, offset);
+    ESP_RETURN_ON_ERROR(wl_erase_range(s_storage_handle->wl_handle, src_addr, size), TAG, "Failed to erase");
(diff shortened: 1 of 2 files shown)


device/esp_tinyusb/tusb_msc_storage.c
```diff
@@ -219,16 +224,12 @@ static esp_err_t _msc_storage_write_sector(uint32_t lba,
         return ESP_ERR_INVALID_STATE;
     }
     size_t sector_size = tinyusb_msc_storage_get_sector_size();
-    size_t temp = 0;
-    size_t addr = 0; // Address of the data to be read, relative to the beginning of the partition.
-    ESP_RETURN_ON_FALSE(!__builtin_umul_overflow(lba, sector_size, &temp), ESP_ERR_INVALID_SIZE, TAG, "overflow lba %lu sector_size %u", lba, sector_size);
-    ESP_RETURN_ON_FALSE(!__builtin_uadd_overflow(temp, offset, &addr), ESP_ERR_INVALID_SIZE, TAG, "overflow addr %u offset %lu", temp, offset);
 
-    if (addr % sector_size != 0 || size % sector_size != 0) {
+    if (size % sector_size != 0) {
```
roma-jam, line 228: To reviewers: 

This is also not relevant and will be deleted in v.2 for two reasons:
- we already have `assert(bufsize <= MSC_STORAGE_BUFFER_SIZE)` on the previous step, which partially covers that
- the MSC driver is using SCSI commands which operates with sectors. So the smaller amount of data that could come from the driver is one sector.

peter-marcisovsky (approved):
Nice fix, LGTM. Is there (theoretical) limit for the maximum capacity of sdmmc, which can be used with esp32?  I am thinking, that we could provide this information into the documentation (if there is a limit), or event to the release notes

tore-espressif (approved):
Very nice fix, thank you @roma-jam !

PR #195, opened 2025-07-01
refactor(tinyusb): Default config header rename

roma-jam:
## Description

When I have added tinyusb_default_configs.h file, I supposed that this file will hold different default configurations for tinyusb.

After implementation, there is only one default configuration, so there is no more point to keep the "s" in the name of the file. 

So, the change is only one across all files:

`tinyusb_default_configs.h` -> `tinyusb_default_config.h`

## Related

- Added in 

## Testing

---

## Checklist


roma-jam:
@peter-marcisovsky 
Thanks, in the documentation I already used the right name from the beginning.

PR #207, opened 2025-07-15
feature(tinyusb_msc): Support second LUN

roma-jam:
## Testing

---

## Checklist

Before submitting a Pull Request, please ensure the following:

- [ ] 🚨 This PR does not introduce breaking changes.
- [ ] All CI checks (GH Actions) pass.
- [ ] Documentation is updated as needed.
- [ ] Tests are updated or added as necessary.
- [ ] Code is well-commented, especially in complex areas.
- [ ] Git history is clean — commits are squashed to the minimum necessary.

PR #208, opened 2025-07-16
refactor(tinyusb_msc): Changed the name of the auto mount flag

roma-jam:
## Description

Changing the flag name to make it common for every storage, not only LUN0.

## Related

N/A

## Testing

N/A

---

## Checklist

### DIFF
--- a/device/esp_tinyusb/include/tinyusb_msc.h
+++ b/device/esp_tinyusb/include/tinyusb_msc.h
@@ -117,12 +117,12 @@ typedef struct {
 typedef struct {
     union {
         struct {
-            uint16_t lun0_auto_mount_off: 1;
+            uint16_t auto_mount_off: 1;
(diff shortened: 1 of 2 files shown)


roma-jam:
@peter-marcisovsky @igi540 

This is a small but important change to the public API: flag now applies for all storages, not only the one which mapped to LUN0. 

I added that flag before and thought to make individual flag for every storage, but after implementing the several storages support in here: https://github.com/espressif/esp-usb/pull/207, I realized that this is not important. 

So, I want to change it before the first release, to not to create the breaking change after. 

PTAL.

roma-jam:
@igi540 

No, no specific requests. Just when I refactored the MSC driver, I saw the possibility to expose SPI Flash and SD/MMC filesystems to USB Host simultaneously with not so much effort. So, I decided to add it.

PR #216, opened 2025-07-25
feature(cdc/modem): Add support for uninstall usb_dte (IEC-337)

ShunzDai:
feature(cdc/modem): Add support for uninstall usb_dte

### DIFF
--- a/host/class/cdc/esp_modem_usb_dte/esp_modem_usb.cpp
+++ b/host/class/cdc/esp_modem_usb_dte/esp_modem_usb.cpp
@@ -36,10 +36,19 @@ static void usb_host_task(void *arg)
 {
     while (1) {
         uint32_t event_flags;
-        usb_host_lib_handle_events(portMAX_DELAY, &event_flags);
+        esp_err_t err = ESP_OK;
+        err = usb_host_lib_handle_events(portMAX_DELAY, &event_flags);
         if (event_flags & USB_HOST_LIB_EVENT_FLAGS_NO_CLIENTS) {
-            ESP_LOGD(TAG, "No more clients: clean up\n");
-            usb_host_device_free_all();
+            err = usb_host_device_free_all();
+            ESP_LOGD(TAG, "No more clients: clean up %d", err);
+            err = usb_host_uninstall();
+            ESP_LOGD(TAG, "USB Host uninstalled %d", err);
+            *(TaskHandle_t *)arg = nullptr;
+            vTaskDelete(NULL);
+        }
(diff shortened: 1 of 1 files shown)


host/class/cdc/esp_modem_usb_dte/esp_modem_usb.cpp
```diff
@@ -36,10 +36,19 @@ static void usb_host_task(void *arg)
 {
     while (1) {
         uint32_t event_flags;
-        usb_host_lib_handle_events(portMAX_DELAY, &event_flags);
+        esp_err_t err = ESP_OK;
+        err = usb_host_lib_handle_events(portMAX_DELAY, &event_flags);
         if (event_flags & USB_HOST_LIB_EVENT_FLAGS_NO_CLIENTS) {
-            ESP_LOGD(TAG, "No more clients: clean up\n");
-            usb_host_device_free_all();
+            err = usb_host_device_free_all();
```
roma-jam, line 42: `usb_host_lib_handle_events` will return `ESP_ERR_NOT_FINISHED` if there are still some device, opened by clients or external hubs are connected. 

So that means, that `usb_host_uninstall` return `ESP_ERR_INVALID_STATE`.

If we want to uninstall the library, we need to release all devices, wait for `USB_HOST_LIB_EVENT_FLAGS_ALL_FREE` flag if there were some devices opened and uninstall the library. 

The good example, how to manage that could be found in the `esp-idf,` in general `usb_host_lib` example: 
https://github.com/espressif/esp-idf/blob/master/examples/peripherals/usb/host/usb_host_lib/main/usb_host_lib_main.c#L126

roma-jam (commented):
Hi @ShunzDai , 

thanks a lot for collaboration! 

Could you please clarify a bit the goal of the changes? 

If the goal is to uninstall the USB Host library, when all the clients are deregistered, then the logic should be a bit different (as we have in esp-idf example).

tore-espressif:
Thank you for the PR @ShunzDai !

The ways of managing life time of USB Host library are pretty diverse beacuse it is a shared system resource. That is why it `esp_modem_usb_dte` the implementation is very naive.

In case you need advanced configuration you can follow advice in this comment: https://github.com/espressif/esp-usb/blob/ff739ce0d7a5cbbdfb7b4f3fa0e8e6e76f0c6aaa/host/class/cdc/esp_modem_usb_dte/esp_modem_usb.cpp#L31

It will be better for you _not to_ instal USB Host Library in `esp_modem_usb_dte` but do it in your application where you will have full control of it

PR #233, opened 2025-08-26
fix(uvc/example): Update camera_display example to IDFv6.0

tore-espressif:


.build-test-rules.yml
```diff
@@ -8,7 +8,7 @@ host/class:
 
 host/class/uvc/usb_host_uvc/examples/camera_display:
   enable:
-    - if: (IDF_VERSION_MAJOR >= 5 and IDF_VERSION_MINOR >= 3)
+    - if: (IDF_VERSION >= "5.3.0")
```
peter-marcisovsky, line 11:
```suggestion
    - if: IDF_VERSION >= 5.3
```

Works locally for me

PR #247, opened 2025-09-05
feat(ci): Run host tests with pytest

peter-marcisovsky:
## Description

Running host tests using `pytest --target-linux -m host_test` instead of the currently used `run_host_tests.sh` shell script.

- added new pytest env marker `host_test`
- added `pytest_*_linux.py` files to each `host_test` folder, so the pytest call can collect them

## Testing

---

## Checklist


peter-marcisovsky:
@igi540  Thank you for the review. Timeout is set to 5 seconds for the whole test app. There is (I think) 7 Linux host tests apps, which are run on a Linux runner. So 35 seconds (as a worst case) for the run of all the test apps is very safe number, considering the build of an test app takes in order of ones of minutes. 

Thus I don't expect any issues with the 5 second timeout for the test run.

PR #271, opened 2025-09-30
Fixed issues with unused variable warnings when compiling with NDEBUG

tore-espressif:
This change was added to esp-idf/usb component after we released 1.0.0 usb.

https://github.com/espressif/esp-idf/commit/484d2c5c2f59ec4432658af0f9192925c39901c6


peter-marcisovsky (approved):
I remember rebasing to this changes. Good catch.

Issue #279, opened 2025-10-05
uvc_host not support high speed on esp32p4? (IEC-394)


Dandjinh:
### Answers checklist.

- [x] I have read the component documentation [ESP-IDF Components](https://components.espressif.com) and the issue is not addressed there.
- [x] I am using target and esp-idf version as defined in component's idf_component.yml
- [x] I have searched the [issue tracker](https://github.com/espressif/esp-usb/issues?q=is%3Aissue) for a similar issue and not found any related issue.

### Which component are you using? If you choose Other, provide details in More Information.

host/uvc

### ESP-IDF version.

v5.5.1

### Development Kit.

ESP32-P4-Module-DEV-KIT

### Used Component version.

v2.3.1

### More Information.

I am trying to get video streaming from ms2130 which converts HDMI streaming to uvc streaming.
I get yuv2 streambing with only 10fps and mjpeg 30fps but only 1/3 of them could be decoded by jpeg-2d.
Looks like only USB-FS bandwidth works.

project uploads here:[esp32p4_uvc.zip](https://github.com/user-attachments/files/22710048/esp32p4_uvc.zip)

The logs folder contains the log of yuv2 mode, the log of mjpeg mode, the description got by `uvc_host_desc_print` and the description got by `UsbTreeView` on Win10.

The result of `uvc_host_desc_print` cannot show any intf whose MPS is bigger than 1024.

Did I miss something to configure USB in HS mode?

tore-espressif:
@Dandjinh Looking at the logs, the camera was correctly enumerated as High Speed. We tested P4 with USB High speed cameras, with maximum MPS (1024) including multiple packet per transaction.

To configure your hardware for better USB receive performance (typical for USB cameras) set `(Top) → Component config → USB-OTG → Hardware FIFO size biasing ` in menconfig to `Bias IN`. This will allocate more space for USB RX buffer. At the same time make sure that when install USB host lib, set `usb_host_config_t.fifo_settings_custom` to zero.

The JPEG decoder is capable of decoding FullHD images at 30fps
https://docs.espressif.com/projects/esp-idf/en/stable/esp32p4/api-reference/peripherals/jpeg.html#jpeg-decoder-performance

Dandjinh:
Hi @tore-espressif , actually the usb device supports MPS 3072.
However, like what I said, MPS more than 1024 cannot be shown under `uvc_host_desc_print`.

[...]

PR #282, opened 2025-10-08
fix(usb): Print additional transaction in microframe

tore-espressif:
High Speed Isochronous and Interrupt endpoints, we also take additional transaction in microframe into account when printing endpoint descriptor.

Related to https://github.com/espressif/esp-usb/issues/279


peter-marcisovsky (approved):
Leaving one suggestion. Also, it would be worth mentioning this in changelog.

PR #284, opened 2025-10-08
feat(type_c,pd): add public APIs for Type-C (HUSB320) and PD (FUSB302)

igi540:
Introduce initial, beta public headers aligned with esp-idf/esp-usb style:
- Type-C core (CC-only): attach/detach/orientation & power-role control
  - File: esp-usb/type_c/include/esp_typec.h
  - Install/uninstall + library info
  - Events: ATTACHED/DETACHED/ORIENTATION/PWR_ROLE/ERROR
  - Factory: esp_typec_port_create_husb320() (HUSB320; CC-only)
  - APIs: esp_typec_set_power_role(), esp_typec_get_orientation(), esp_typec_is_attached(), esp_typec_port_destroy()

- Power Delivery (PD-only): runtime role control and sink requests
  - File: esp-usb/type_c/include/esp_typec_pd.h
  - Install/uninstall + optional polling handler + library info
  - Events: ATTACHED/DETACHED/CONTRACT_READY/CONTRACT_LOST/PPS_ADJUSTED/ERROR
  - Factory: esp_typec_pd_port_create_fusb302() (FUSB302; PD-capable)
  - APIs: esp_typec_pd_set_power_role(), esp_typec_pd_sink_request_fixed(), esp_typec_pd_sink_request_pps(), esp_typec_pd_get_contract(), esp_typec_pd_get_orientation(), esp_typec_pd_is_attached(), esp_typec_pd_port_destroy()

Decisions:
- Split CC-only and PD APIs to keep modules independent and build-light.
- Reused event naming across CC and PD for consistency; separate enums to avoid coupling.
- Only include HUSB320 (CC-only) and FUSB302 (PD-capable) factories for now; removed any HUSB311 references.
- New files use SPDX year 2025 per repo convention.

Notes:
- APIs are intentionally minimal; more controls (source caps, swaps, DRP policies) can be added without breaking the current surface.
- Data-role (host/device) integration is out of scope for this step.

## Testing

---

## Checklist


peter-marcisovsky:
FYI, to fix the failing linux host test, update the `.build-test-rules.yml` in the root dir with something like:

```yml
type_c:
  disable:
    - if: IDF_TARGET in ["linux"]

type_c_manager:
  enable:
    - if: SOC_USB_OTG_SUPPORTED == 1
 ```
 
Since the `type_c` component does not depend on `USB_OTG` peripheral, it will be usable with any target.. if I remember correctly.
 
And the `type_c_manager` will work directly with `usb` and `esp_tinyusb` components.
 
The `idf-build-apps find --target linux` in `build_and_run_host_test.yml` searches for buildable apps in the whole repo. Since the newly added `type_c` and `type_c_manager` build rules have not been set yet, the `..host_test.yml` is trying to build it for linux target.

File: type_c/examples/pd_fusb302/main/CMakeLists.txt [hunk 1]
```diff
@@ -0,0 +1,5 @@
+idf_component_register(
+    SRCS "main.c"
+    PRIV_REQUIRES esp_driver_i2c esp_driver_gpio
+    REQUIRES type_c
```
peter-marcisovsky: Not that it would not be correct, but in all the examples and test apps, we specify the local components using `idf_component.yml` file instead of the cmake. 

Could you please move the `type_c` dependency to the `idf_component.yml` using `override_path` to the local component is used fro build, not managed.

type_c/examples/pd_fusb302/sdkconfig.defaults
```diff
@@ -0,0 +1 @@
+CONFIG_LOG_DEFAULT_LEVEL_INFO=y
```
peter-marcisovsky: I believe this is a default config anyway, you don't have to explicitly specify it.

type_c/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+idf_component_register(
+    SRCS "src/esp_typec.c" "src/fusb302_ctrl.c"
```
peter-marcisovsky: You are not sourcing the other 2 `.c` files from the `src/` here.

type_c/README.md
```diff
@@ -0,0 +1,110 @@
...
+| ---------: | ---------- | --------------------------------- | --------------------------------- |
+|    FUSB302 | Type-C CC  | `esp_typec_port_create_fusb302()` | CC attach/detach and role control |
+
+______________________________________________________________________
+
+## Quick Start
+
+### Type-C CC with FUSB302
+
+```c
```
peter-marcisovsky: This is just copy-pasted example, quick start should be briefer. For example [here](https://github.com/espressif/esp-usb/tree/master/host/usb#usage)

.build-test-rules.yml
```diff
@@ -28,6 +28,12 @@ host/class/msc:
   enable:
     - if: SOC_USB_OTG_SUPPORTED == 1
 
+type_c:
+  depends_filepatterns:
+    - 'type_c/**'
+  disable:
+    - if: IDF_TARGET in ["linux"]
```
peter-marcisovsky: There should be Linux and also targets which do not feature I2C driver.

peter-marcisovsky (commented):
Leaving some "housekeeping" points. Will check the actual code later.

- Q1: Are you planning to add some tests to this PR?
- Q2: If yes for tests, are you planning to add some CI run for this component? As right now there is at leas the example, which could be run in CI
- Q3: Are you planning to add documentation to this PR?

type_c/examples/pd_fusb302/main/main.c
```diff
@@ -0,0 +1,101 @@
...
+    default:
+        ESP_LOGW(TAG, "EVENT %d", (int)id);
+        break;
+    }
+}
+
+/* ===== App entry ===== */
+void app_main(void)
+{
+    ESP_LOGI("main_task", "Calling app_main()");
```
peter-marcisovsky: Isn't this log already included into the bootloader log?

This is an output from the example examples
``` shell
I (320) main_task: Started on CPU0
I (330) main_task: Calling app_main()
I (330) main_task: Calling app_main()
```

type_c/examples/pd_fusb302/main/main.c
```diff
@@ -0,0 +1,101 @@
...
+    /* --- Create typec port (FUSB302 backend) --- */
+    esp_typec_port_handle_t port = NULL;
+
+    esp_typec_port_config_t port_cfg = {
+        .default_power_role = ESP_TYPEC_PWR_SINK,
+        .task_stack = 4096,
+        .task_prio  = 5,
+        .src_vbus_gpio   = VBUS_GPIO_PH,
+        .src_vbus_gpio_n = VBUS_GPIO_N,
+        .rp_current = ESP_TYPEC_RP_1A5,
```
peter-marcisovsky: Just to reorganize the struct members according to the struct typedef. I think cpp build would produce an error here.

```suggestion
        .default_power_role = ESP_TYPEC_PWR_SINK,
        .task_stack = 4096,
        .task_prio  = 5,
        .rp_current = ESP_TYPEC_RP_1A5,
        .src_vbus_gpio   = VBUS_GPIO_PH,
        .src_vbus_gpio_n = VBUS_GPIO_N,
```

type_c/src/fusb302_ctrl.c
```diff
@@ -0,0 +1,673 @@
...
+
+    if (is_source) {
+        /* Advertise the configured Rp current after attach. */
+        ESP_RETURN_ON_ERROR(fusb302_set_host_cur(dev, rp_current_to_host_cur_bits(dev->rp_current)),
+                            TAG, "set HOST_CUR attach");
+        ESP_RETURN_ON_ERROR(i2c_wr8(dev->hw.i2c_dev, REG_MEASURE,
+                                    rp_current_to_mdac(dev->rp_current)),
+                            TAG, "set MDAC attach");
+        /* Debounce to avoid false detach during the first few ms after attach. */
+        vTaskDelay(pdMS_TO_TICKS(SRC_ATTACH_DEBOUNCE_MS));
```
peter-marcisovsky: The default freertos tick is 10ms, any delay shorter than that is just rounded to 10ms. If you really need 3ms not 10ms, you can use busy wait or timer.

type_c/src/esp_typec.c
```diff
@@ -0,0 +1,702 @@
...
+    ESP_GOTO_ON_ERROR(ctx->backend->service_irq(ctx->ctrl, &ev), fail, TAG, "prime irq");
+
+    /* Now arm the GPIO */
+    gpio_intr_enable(ctx->gpio_int);
+
+    *port_hdl_ret = (esp_typec_port_handle_t)ctx;
+    ESP_LOGI(TAG, "Type-C task started");
+    return ESP_OK;
+fail:
+    ESP_LOGE(TAG, "failed, err=%d", ret);
```
peter-marcisovsky: Or something similar. 
```suggestion
    ESP_LOGE(TAG, "Init failed, err=%d", esp_err_to_name(ret));
```

igi540:
@peter-marcisovsky @tore-espressif Updates are in — ready for another round of review. Thanks!

type_c/usb_tcpm/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+1. Install the TCPM library with `usb_tcpm_install()`
+2. Create a port using `usb_tcpm_port_create_fusb302()`
+3. Register for `USB_TCPM_EVENT` events via `esp_event_handler_register()`
+4. (Optional) Read a snapshot with `usb_tcpm_get_status()`
+5. Destroy the port with `usb_tcpm_port_destroy()`
+
+## Examples
+
+- `examples/pd_fusb302` (logs attach/detach events and prints a one-shot status snapshot after startup)
```
peter-marcisovsky: Just to link the path to the example. Please make sure if the path I suggested is correct.
```suggestion
- [pd_fusb302](../examples/pd_fusb302) (logs attach/detach events and prints a one-shot status snapshot after startup)
```

type_c/usb_tcpm/test_apps/pd_fusb302/sdkconfig.defaults
peter-marcisovsky: Please add the default esp32p4 config to set ECO  version for build, same as it is done in the rest of the test [apps](https://github.com/espressif/esp-usb/blob/master/host/class/cdc/usb_host_cdc_acm/test_app/sdkconfig.defaults.esp32p4)

type_c/usb_tcpm/examples/pd_fusb302/main/main.c
```diff
@@ -0,0 +1,118 @@
...
+        ESP_LOGI(TAG, "status: attached=%d cc2=%d rp=%u mA role=%d",
+                 st.attached, st.cc2_active, (unsigned)st.rp_cur_ma, (int)st.role);
+    }
+
+    /* App can idle; the Type-C task & ISR do the work */
+    while (true) {
+        vTaskDelay(pdMS_TO_TICKS(1000));
+    }
+
+    /* (Optional cleanup if you ever leave the loop)
```
peter-marcisovsky: You can leave the clean up uncommented. Is't not reachable code anyway.

type_c/usb_tcpm/src/fusb302_ctrl.c
```diff
@@ -0,0 +1,824 @@
...
+
+// Write 8-bit value to a FUSB302 register via I2C.
+static inline esp_err_t i2c_wr8(i2c_master_dev_handle_t dev, uint8_t reg, uint8_t val)
+{
+    uint8_t buf[2] = { reg, val };
+    return i2c_master_transmit(dev, buf, sizeof(buf), 100 /*ms*/);
+}
+
+// Read 8-bit value from a FUSB302 register via I2C.
+static inline esp_err_t i2c_rd8(i2c_master_dev_handle_t dev, uint8_t reg, uint8_t *val)
```
peter-marcisovsky: You are using `dev` variable for both the `i2c_master_dev_handle_t` and the `fusb302_dev_t` could you differentiate it a bit?

type_c/usb_tcpm/src/fusb302_ctrl.c
```diff
@@ -0,0 +1,824 @@
...
+        return ESP_ERR_INVALID_ARG;
+    }
+    i2c_master_dev_handle_t d = dev->hw.i2c_dev;
+
+    ESP_GOTO_ON_ERROR(fusb302_enable_irq(dev, false), fail, TAG, "disable irq");
+    ESP_GOTO_ON_ERROR(i2c_wr8(d, REG_SWITCHES0, 0x00), fail, TAG, "SWITCHES0 Hi-Z");
+    ESP_GOTO_ON_ERROR(i2c_wr8(d, REG_CONTROL2, CTL2_MODE_NONE), fail, TAG, "CONTROL2 mode none");
+    ESP_GOTO_ON_ERROR(i2c_wr8(d, REG_POWER, PWR_PWR_LOW), fail, TAG, "power down");
+
+fail:
```
peter-marcisovsky: Is there any point of jumping to fail and freeing the dev pointer if any of the above failed? Maybe you can use `ESP_ERROR_CHECK` instead of the `ESP_GOTO_ON_ERROR`. Just a suggestion, I don't know this code well.

type_c/usb_tcpm/src/usb_tcpm.c
```diff
@@ -0,0 +1,698 @@
...
+    if (ctx->cfg.src_vbus_gpio != GPIO_NUM_NC) {
+        gpio_set_level(ctx->cfg.src_vbus_gpio, enable); // active-high
+    }
+    if (ctx->cfg.src_vbus_gpio_n != GPIO_NUM_NC) {
+        gpio_set_level(ctx->cfg.src_vbus_gpio_n, !enable); // active-low
+    }
+    ESP_LOGD(TAG, "VBUS=%d", enable);
+}
+
+static void usb_tcpm_handle_events(typec_port_t *ctx, typec_evt_mask_t ev)
```
peter-marcisovsky: Could you rework this function an make it bit shorter? As right now it has over 300 lines.
Maybe use separate function for each role?

peter-marcisovsky (commented):
Also, you can add:
- a pytest file, once we enable this in CI run
- the type_c folder into the static analysis

type_c/usb_tcpm/idf_component.yml
```diff
@@ -0,0 +1,29 @@
...
+description: ESP-IDF USB Type-C stack (beta)
+version: "0.1.0-beta"
+url: https://github.com/espressif/esp-usb/tree/master/type_c
+issues: "https://github.com/espressif/esp-usb/issues"
+repository: "https://github.com/espressif/esp-usb.git"
+repository_info:
+  path: "type_c"
+dependencies:
+  idf: ">=5.4"
+targets:
```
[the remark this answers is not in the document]
tore-espressif: All our chips support I2C so far, so IMO, removing the 'targets' section is the best way going forward

type_c/usb_tcpm/README.md
peter-marcisovsky: Can skip for beta release

type_c/usb_tcpm/src/fusb302_ctrl.c
```diff
@@ -0,0 +1,824 @@
...
+}
+
+esp_err_t fusb302_init(const fusb302_hw_cfg_t *hw, fusb302_dev_t **out)
+{
+    if (!hw || !out) {
+        return ESP_ERR_INVALID_ARG;
+    }
+
+    esp_err_t ret = ESP_OK;
+    fusb302_dev_t *dev = calloc(1, sizeof(*dev));
```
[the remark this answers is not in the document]
peter-marcisovsky: The comment was more general, since there is dynamic allocation in several places. I did not check sizes all of the structs. We can potentially add this in future once the structs grow in size, or for targets with small internal RAM. 
Skip for now.

peter-marcisovsky (approved):
LGTM

PR #309, opened 2025-11-04
feature(esp_tinyusb): Remote wakeup device test application [WIP]

roma-jam:
## Description

Test application for Remote Wakeup feature on ESP32-S2, ESP32-S3, ESP32-P4 (USB OTG 1.1 and USB OTG 2.0).

Test requires external USB Host, which might be send to sleep mode by an operator. 

Logic of the test is linear, all events (ATTACH, SUSPEND, RESUME) are verified and the test simplifies the verification of the feature: 
- connect the device to the host
- run the test
- initiate the sleep
- get the result

Possible to run on USB1.1 for S2/S3 and on USB2.0 and USB1.1 for P4. 

### ESP32-P4 USB OTG 2.0

image

### ESP32-P4, USB OTG 1.1

image

## Related

N/A

## Testing

Added additional test: 
image

---

## Checklist

### DIFF
--- a/device/esp_tinyusb/test_apps/remote_wakeup/CMakeLists.txt
+++ b/device/esp_tinyusb/test_apps/remote_wakeup/CMakeLists.txt
@@ -1,0 +1,9 @@
+# The following lines of boilerplate have to be in your project's
+# CMakeLists in this exact order for cmake to work correctly
+cmake_minimum_required(VERSION 3.16)
+include($ENV{IDF_PATH}/tools/cmake/project.cmake)
+
+# "Trim" the build. Include the minimal set of components, main, and anything it depends on.
+idf_build_set_property(MINIMAL_BUILD ON)
+
+project(test_app_remote_wakeup)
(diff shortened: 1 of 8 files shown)


tore-espressif (changes requested):
@roma-jam Thank you for preparing this test. Before moving forward, there are several points we should address:

1. Test app structure
Please analyze whether creating a new test_app is truly necessary for these two new test cases. Could they be integrated into an existing test application instead?
The motivation is to minimize build and test duration. Each additional test_app increases build time and significantly prolongs on-target testing (due to new artifacts, separate flashing, and setup). In contrast, extending an existing app with two new test cases would have a negligible impact on both build and execution time.

2. Automation and CI integration
Please add a corresponding pytest_*.py file so these tests can run automatically in CI. This ensures the new functionality is continuously verified and doesn’t rely on manual execution.

3. Reducing operator interaction
The current test procedure still requires considerable manual action, which makes it unsuitable for automated CI runs.
Since the USB Host library already supports the suspend feature and Remote Wake-up support is being added, it would be preferable to adapt this test to run fully automated between two ESP boards connected via USB. This approach aligns with how we already execute USB class driver tests and can reuse the existing dual-runner infrastructure.

tore-espressif:
> Ok, I can keep it locally for now.

No, the goal is to have it in CI, running on our dual target runners

PR #344, opened 2025-12-01
Fix security issues

tore-espressif:
These security issues were flagged and fixed by Claude AI. Based on #326 , that was divided into separate commits.

These issues and their fixes must be validated first. 

### TODO:
- Validate class driver fixes (HID, UAC, MSC) @tore-espressif 
    - [x] HID: Valid, needs code change
    - [x] MSC: Cannot occur. The hardware (USB-OTG) makes sure that it writes only requested amount of data
    - [x] UAC: Valid, but the code should be changed cc @leeebo 
- [x] Validate esp_tinyusb fixes @igi540 
- [x] Validate host lib fixes @peter-marcisovsky 

Closes #326

### DIFF
--- a/host/class/hid/usb_host_hid/CHANGELOG.md
+++ b/host/class/hid/usb_host_hid/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Fixed a vulnerability with overwrite freed heap memory during `hid_host_get_report_descriptor()`
+- Fixed a vulnerability with overwrite freed heap memory in `hid_host_get_report_descriptor()` and `hid_class_request_set()`
 
 ### Added
 
--- a/host/class/hid/usb_host_hid/hid_host.c
+++ b/host/class/hid/usb_host_hid/hid_host.c
@@ -768,6 +768,40 @@ static esp_err_t hid_control_transfer(hid_device_t *hid_device,
 }
 
 /**
+ * @brief Reallocate control transfer buffer if needed for larger size
+ *
+ * @param[in] hid_device     Pointer to HID device structure
+ * @param[in] required_size  Required buffer size (must include USB_SETUP_PACKET_SIZE)
+ * @note Device must be locked before calling this function
+ * @return
+     - ESP_OK on success
+     - ESP_ERR_NO_MEM on allocation failure
+     - ESP_ERR_INVALID_SIZE on invalid size
+ */
+static esp_err_t hid_device_realloc_ctrl_xfer(hid_device_t *hid_device, size_t required_size)
+{
+    ESP_RETURN_ON_FALSE(required_size <= HID_MAX_REPORT_DESC_LEN, ESP_ERR_INVALID_SIZE, TAG, "Requested EP0 transfer buffer size exceeds maximum");
+
+    const size_t ctrl_size = hid_device->ctrl_xfer ? hid_device->ctrl_xfer->data_buffer_size : 0;
+
+    if (ctrl_size < required_size) {
+        ESP_LOGD(TAG, "Change HID ctrl xfer size from %"PRIu32" to %"PRIu32"",
+                 (uint32_t) ctrl_size,
+                 (uint32_t) required_size);
+
+        usb_host_transfer_free(hid_device->ctrl_xfer);
+
+        if (usb_host_transfer_alloc(required_size, 0, &hid_device->ctrl_xfer) != ESP_OK) {
+            ESP_LOGE(TAG, "Unable to re-allocate transfer buffer for EP0");
+            hid_device->ctrl_xfer = NULL;
+            return ESP_ERR_NO_MEM;
+        }
+    }
+
+    return ESP_OK;
+}
+
+/**
  * @brief USB class standard request get descriptor
  *
  * @param[in] hid_device  Pointer to HID device structure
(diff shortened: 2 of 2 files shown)


device/esp_tinyusb/CMakeLists.txt
```diff
@@ -61,3 +61,18 @@ endif()
 # Pass tusb_config.h from this component to TinyUSB
 idf_component_get_property(tusb_lib ${tinyusb_name} COMPONENT_LIB)
 target_include_directories(${tusb_lib} PRIVATE "include")
+
+# Add security hardening compiler flags
+target_compile_options(${COMPONENT_LIB} PRIVATE
```
peter-marcisovsky: Should we add similar (or the same) compiler flags to all the component cmakes? To have unified build type in the repo?

host/usb/src/enum.c
```diff
@@ -368,6 +377,14 @@ static void control_request_general(enum_stage_t stage)
...
     case ENUM_STAGE_GET_FULL_CONFIG_DESC: {
         // Get the full configuration descriptor at descriptor index, requesting its exact length.
+        // Bounds check with overflow protection
+        uint16_t total_transfer_size = 0;
+        if (__builtin_add_overflow(sizeof(usb_setup_packet_t), wTotalLength, &total_transfer_size) ||
+                total_transfer_size > ENUM_CTRL_TRANSFER_MAX_DATA_LEN) {
+            ESP_LOGE(ENUM_TAG, "Configuration descriptor size overflow (wTotalLength=%u)", wTotalLength);
+            ret = ESP_ERR_INVALID_SIZE;
+            goto exit;
+        }
```
peter-marcisovsky: This does not seem like a complete solution: `ret` and `goto` are not defined. Anyway it seems, that it is not needed anyway, since thehere is the same check in `parse_short_config_desc`

```
#if (ENUM_CTRL_TRANSFER_MAX_DATA_LEN < UINT16_MAX)  // Suppress -Wtype-limits warning due to uint16_t wTotalLength
    // Check if the descriptor is too long to be supported
    if (config_desc->wTotalLength > ENUM_CTRL_TRANSFER_MAX_DATA_LEN) {
        ESP_LOGE(ENUM_TAG, "Configuration descriptor larger than control transfer max length");
        ret = ESP_ERR_INVALID_SIZE;
        goto exit;
    }
#endif
```

host/usb/src/enum.c
```diff
@@ -449,9 +468,11 @@ static esp_err_t parse_short_dev_desc(void)
...
     const usb_device_desc_t *dev_desc = (usb_device_desc_t *)(ctrl_xfer->data_buffer + sizeof(usb_setup_packet_t));
 
-    // Check if the returned descriptor has correct type
-    if (dev_desc->bDescriptorType != USB_B_DESCRIPTOR_TYPE_DEVICE) {
-        ESP_LOGE(ENUM_TAG, "Short dev desc has wrong bDescriptorType");
+    // Check if the returned descriptor has correct type and length
+    if (dev_desc->bDescriptorType != USB_B_DESCRIPTOR_TYPE_DEVICE ||
+            dev_desc->bLength != sizeof(usb_device_desc_t)) {
+        ESP_LOGE(ENUM_TAG, "Short dev desc invalid (type=%d, len=%d, expected len=%d)",
+                 dev_desc->bDescriptorType, dev_desc->bLength, sizeof(usb_device_desc_t));
```
peter-marcisovsky: OK:

USB spec compliance: The device descriptor has a fixed length (18 bytes in USB 2.0). If bLength is wrong, the descriptor is malformed.


TODO:
- [ ] Check that bLength >= sizeof(usb_device_desc_t) instead of ==. Some stacks allow extensions, though the USB spec defines a fixed length.
- [] Consider validating bMaxPacketSize0 against allowed values (8, 16, 32, 64 for full‑speed devices).

host/usb/src/enum.c
```diff
@@ -497,9 +518,11 @@ static esp_err_t parse_full_dev_desc(void)
...
     const usb_device_desc_t *dev_desc = (usb_device_desc_t *)(ctrl_xfer->data_buffer + sizeof(usb_setup_packet_t));
 
-    // Check if the returned descriptor has correct type
-    if (dev_desc->bDescriptorType != USB_B_DESCRIPTOR_TYPE_DEVICE) {
-        ESP_LOGE(ENUM_TAG, "Full dev desc has wrong bDescriptorType");
+    // Check if the returned descriptor has correct type and length
+    if (dev_desc->bDescriptorType != USB_B_DESCRIPTOR_TYPE_DEVICE ||
+            dev_desc->bLength != sizeof(usb_device_desc_t)) {
+        ESP_LOGE(ENUM_TAG, "Full dev desc invalid (type=%d, len=%d, expected len=%d)",
+                 dev_desc->bDescriptorType, dev_desc->bLength, sizeof(usb_device_desc_t));
```
peter-marcisovsky: Same as above

host/usb/src/usb_helpers.c
```diff
@@ -23,6 +26,14 @@ const usb_standard_desc_t *usb_parse_next_descriptor(const usb_standard_desc_t *
     if (*offset >= wTotalLength) {
         return NULL;    // We have traversed the entire configuration descriptor
     }
+
+    // Validate bLength before using it for pointer arithmetic
+    // Descriptors must be at least 2 bytes and have even length for proper alignment
+    if (cur_desc->bLength < 2 || (cur_desc->bLength & 1)) {
+        ESP_LOGW(TAG, "Invalid descriptor length: %d (must be >=2 and even)", cur_desc->bLength);
+        return NULL;
+    }
```
peter-marcisovsky: Ok.

`cur_desc->bLength` is used directly for pointer arithmetic and offset updates.

If bLength is 0, you’ll never advance → infinite loop.
If bLength is 1, pointer arithmetic misaligns.
If bLength is odd or corrupted, you can misinterpret memory or walk off the buffer.

- Must be at least 2 bytes (minimum descriptor header size).
- Must be even (USB descriptors are word‑aligned).

host/usb/src/usb_helpers.c
```diff
@@ -278,7 +297,17 @@ void usb_print_config_descriptor(const usb_config_desc_t *cfg_desc, print_class_
...
+#define MAX_DESCRIPTOR_ITERATIONS 256
+    int iteration_count = 0;
+
     do {
+        // Check iteration limit
+        if (++iteration_count > MAX_DESCRIPTOR_ITERATIONS) {
+            ESP_LOGW(TAG, "Too many descriptors (>%d), aborting parse", MAX_DESCRIPTOR_ITERATIONS);
+            break;
+        }
+
```
peter-marcisovsky: Ok,

Risks:
- If bLength fields are corrupted (e.g. zero length or looping back), usb_parse_next_descriptor could keep returning non‑NULL indefinitely.
- That leads to an infinite loop or excessive iterations, potentially consuming CPU or stack resources.

Prevents infinite loops or runaway parsing caused by malformed descriptors.

PR #355, opened 2025-12-15
fix(host/usb): Fixed deadlock that prevented closing devices from hig…

tore-espressif:
If in usb_host_interface_release() the underlying pipe was halted while having an URB in-flight,
usb_host_client_handle_events() may not have had a chance to process the URB yet.
In this case we retry after 10 FreeRTOS ticks to give the class driver task chance to run.

Closes https://github.com/espressif/esp-idf/issues/17707

### DIFF
--- a/host/class/cdc/usb_host_cdc_acm/include/usb/cdc_acm_host.h
+++ b/host/class/cdc/usb_host_cdc_acm/include/usb/cdc_acm_host.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2015-2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2015-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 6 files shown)


peter-marcisovsky (commented):
Leaving one thought. 
Also, I was running the test app with esp32s2 and the test was failing with too much leaked memory.. don't know if it's relevant to the issue, or to the test case.

```bash
Running closing_high_priority...
USB Host installed
Installing CDC-ACM driver with minimum priority
Opening CDC-ACM device
MALLOC_CAP_8BIT usage: Free memory delta: -24628 Leak threshold: -30 
MALLOC_CAP_8BIT critical leak: Before 243896 bytes free, After 219268 bytes free (delta 24628)
./main/test_cdc_acm_host.cpp:42:closing_high_priority:FAIL:Function [cdc_acm].  The test leaked too much memory
Test ran in 1060ms

-----------------------
1 Tests 1 Failures 0 Ignored 
FAIL
```

host/usb/CMakeLists.txt
```diff
@@ -18,8 +18,11 @@ if(${target} STREQUAL "esp32p4")
 endif()
 
 if(${IDF_VERSION_MAJOR} LESS 6)
-    # usb_phy driver is present in USB component only in IDF 5.x
-    list(APPEND priv_requires esp_driver_gpio)      # usb_phy driver relies on gpio driver API
+    # usb_phy driver is present in USB component only in IDF 5.x, it relies on gpio driver API
+    list(APPEND priv_requires esp_driver_gpio)
+else()
+    # For IDF 6.x and later, usb_phy driver is moved to esp_hw_support and HAL driver is moved to esp_hal_usb
+    list(APPEND priv_requires esp_hal_usb esp_hw_support)
```
peter-marcisovsky: Has something changed wit the private components? Or is it just a good practice to append this components? Because the build has been working event without this change.

tore-espressif: No, nothing has changed now. All esp_hal_** components are added as dependncy by default.

However, this can change in future, where only required HAL components are build and linked. With this change we will be one step ahead

PR #368, opened 2026-01-08
CI: Fix conditional builds

tore-espressif:
## Description

With collect_app_info we gather more detailed information about which apps were built. This information is later used in run jobs.

## Related

- Fixes #359 
- No apps build; no apps run job https://github.com/espressif/esp-usb/actions/runs/20817646590
- UAC build; no apps run job https://github.com/espressif/esp-usb/actions/runs/20818540798
- all apps build and run https://github.com/espressif/esp-usb/actions/runs/20819010563
## Testing

CI only

---

## Checklist

### DIFF
--- a/.github/workflows/build_and_run_test_app_usb.yml
+++ b/.github/workflows/build_and_run_test_app_usb.yml
@@ -79,10 +79,21 @@ jobs:
           fi
 
           idf-build-apps find --paths ${{ matrix.path }} --disable-targets linux $MODIFIED_FILES_FLAG
-          idf-build-apps build --paths ${{ matrix.path }} --disable-targets linux $MODIFIED_FILES_FLAG
+          idf-build-apps build --paths ${{ matrix.path }} --disable-targets linux $MODIFIED_FILES_FLAG --collect-app-info="app_info_${{ github.run_id }}.txt"
+      - name: Check if build artifacts exist
+        id: check_build_artifacts
+        shell: bash
+        run: |
+          if find . -path "*/build_esp*/*" -type f -print -quit | grep -q .; then
+            echo "build_artifacts_exist=true" >> $GITHUB_OUTPUT
+          else
+            echo "build_artifacts_exist=false" >> $GITHUB_OUTPUT
+          fi
       - uses: actions/upload-artifact@v4
+        if: steps.check_build_artifacts.outputs.build_artifacts_exist == 'true'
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_run_test_app_usb.yml
```diff
@@ -90,7 +100,7 @@ jobs:
             **/build_esp*/test_app_*.elf
             **/build_esp*/flasher_args.json
             **/build_esp*/config/sdkconfig.json
-          if-no-files-found: ignore # Ignore if no builds were run. This is possible since we have conditional builds.
```
peter-marcisovsky: Should we return the `if-no-files-found: error` now? Just to double check the file presence for the attifacts upload action.

.idf_ci.toml
```diff
@@ -0,0 +1,20 @@
+component_mapping_regexes = [
+    "/host/class/msc/(.+)/",
+    "/host/class/cdc/(.+)/",
+    "/host/class/uac/(.+)/",
+    "/host/class/uvc/(.+)/",
+    "/host/class/hid/(.+)/",
```
peter-marcisovsky: What about `usb` and `esp_tinyusb` components? Just guessing here.. I have not configured anything related `idf_ci` before.

.idf_ci.toml
```diff
@@ -0,0 +1,20 @@
...
+    "/common_components/(.+)/",
```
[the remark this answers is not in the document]
tore-espressif: Yes, copy pasted. I'll investigate its purpose and update accordingly

.idf_ci.toml
tore-espressif: added

.idf_ci.toml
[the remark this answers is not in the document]
tore-espressif: @peter-marcisovsky thank you for the review!

These component mapping regexes are used to map changed files to changed components. This can be used during build step with idf-build-apps.

Instead of using `depends_filepatters` in .build-test-rules.yml we can then use `depends_componets` which is cleaner and easier to maintain.

It also has the advantage on trigerring the build&test only on _code_ related changes. Changed files that match `component_ignored_file_extensions` do not trigger the conditional build. So eg README.md change does not trigger all test build&test. We can implement this in next steps of improving CI, if current approach proves lacking

PR #389, opened 2026-01-26
feat(hub): Added support for multiple root ports

tore-espressif:
This will allow having both FS and HS root ports active at the same time and connected devices working properly.

---

> [!NOTE]
> **Medium Risk**
> Touches core USB host hub/enumeration plumbing and root-port state management, which can affect device attach/detach, enumeration ordering, and suspend/resume behavior across ports. Changes are scoped but concurrency/ISR-driven paths and multi-port edge cases increase regression risk.
> 
> **Overview**
> **Adds multi-root-port support across the USB host stack.** The Hub driver now initializes/powers multiple HCD root ports based on `hub_config->port_map`, tracks per-port state/requests, and routes ISR events and recovery/disable actions per root port.
> 
> **Makes devices root-port aware end-to-end.** Device tree nodes and `usbh_devs_add()` now associate devices with a specific `root_port_hdl`, adds `usbh_dev_get_root_port_hdl()` and `ext_hub_get_root_port()`, and tightens the “address 0” constraint to only conflict within the same root port.
> 
> **Improves enumeration scheduling for multi-port scenarios.** Enumeration starts by UID via new `usbh_devs_open_uid()`, and adds a small pending-UID queue so concurrent attach events enqueue and enumerate sequentially (returning `ESP_ERR_NO_MEM` if the queue is full).
> 
> **Test/build updates.** Moves/standardizes test port/PHY selection macros, adjusts an isoch test to use `TEST_PORT_NUM`, bumps a CDC test leak threshold, and adds a Linux mock define for `SOC_USB_OTG_PERIPH_NUM`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f4564386ef8f958bcf99f876e1659b14c25be635. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/host/usb/src/usbh.c
+++ b/host/usb/src/usbh.c
@@ -162,7 +162,7 @@ static device_t *_find_dev_from_uid(unsigned int uid)
     */
     device_t *dev_iter;
 
-    // Search the device lists for a device with the specified address
+    // Search the device lists for a device with the specified uid
     TAILQ_FOREACH(dev_iter, &p_usbh_obj->dynamic.devs_idle_tailq, dynamic.tailq_entry) {
         if (dev_iter->constant.uid == uid) {
             return dev_iter;
@@ -940,8 +940,9 @@ esp_err_t usbh_devs_add(usbh_dev_params_t *params)
         ret = ESP_ERR_INVALID_ARG;
         goto exit;
     }
-    // Check that there is not already a device currently with address 0
-    if (_find_dev_from_addr(0) != NULL) {
+    // Check that there is not already a device currently with address 0 (not enumerated) on the same port
+    device_t *dev_with_addr_0 = _find_dev_from_addr(0);
+    if (dev_with_addr_0 != NULL && dev_with_addr_0->constant.port_hdl == params->root_port_hdl) {
         ret = ESP_ERR_NOT_FINISHED;
         goto exit;
     }
@@ -1131,14 +1132,12 @@ void usbh_devs_set_pm_actions_all(usbh_dev_ctrl_t device_ctrl)
     }
 }
 
-esp_err_t usbh_devs_open(uint8_t dev_addr, usb_device_handle_t *dev_hdl)
+static esp_err_t _dev_open(device_t *dev_obj)
 {
-    USBH_CHECK(dev_hdl != NULL, ESP_ERR_INVALID_ARG);
+    /*
+    THIS FUNCTION MUST BE CALLED FROM A CRITICAL SECTION
+    */
     esp_err_t ret;
-
-    USBH_ENTER_CRITICAL();
-    // Go through the device lists to find the device with the specified address
-    device_t *dev_obj = _find_dev_from_addr(dev_addr);
     if (dev_obj != NULL) {
         // Check if the device is in a state to be opened
         if (dev_obj->dynamic.flags.is_gone ||           // Device is already gone (disconnected)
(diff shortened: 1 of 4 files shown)


host/usb/src/usbh.c
```diff
@@ -1131,14 +1132,9 @@ void usbh_devs_set_pm_actions_all(usbh_dev_ctrl_t device_ctrl)
     }
 }
 
-esp_err_t usbh_devs_open(uint8_t dev_addr, usb_device_handle_t *dev_hdl)
+esp_err_t _dev_open(device_t *dev_obj)
```
peter-marcisovsky:
```suggestion
static esp_err_t _dev_open(device_t *dev_obj)
```
+ some description, that the function must be called from critical section

host/usb/src/hub.c
```diff
@@ -48,11 +48,13 @@ implement the bare minimum to control the root HCD port.
  */
 typedef enum {
     HUB_DRIVER_ACTION_NONE                = 0,
-    HUB_DRIVER_ACTION_ROOT_EVENT          = (1 << 0),
-    HUB_DRIVER_ACTION_ROOT_REQ            = (1 << 1),
+    HUB_DRIVER_ACTION_ROOT_EVENT          = BIT0,
+    HUB_DRIVER_ACTION_ROOT_REQ            = BIT1,
```
peter-marcisovsky: Do you want to keep this enums without port index? 
We have root without index and root 1:
- `..ACTION_ROOT_EVENT` 
- `..ACTION_ROOT1_EVENT`

host/usb/src/hub.c
```diff
@@ -738,43 +799,88 @@ esp_err_t hub_root_start(void)
...
+    }
+    if (all_stopped) {
         // The HUB was already stopped by usb_host_lib_set_root_port_power(false)
         HUB_DRIVER_EXIT_CRITICAL();
         return ESP_OK;
     }
-    p_hub_driver_obj->dynamic.root_port_state = ROOT_PORT_STATE_NOT_POWERED;
     HUB_DRIVER_EXIT_CRITICAL();
 
     // HCD_PORT_CMD_POWER_OFF will only fail if the port is already powered_off
```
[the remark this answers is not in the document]
peter-marcisovsky: I meant, that the `_port_cmd_power_off` command can fail not only according to the comment:

> // HCD_PORT_CMD_POWER_OFF will only fail if the port is already powered_off

It also can fail during enabling the clock, aka the `_internal_clk_gate` could fail. But that would be a failure of a HW, not a SW.

So my comment was more like: FYI there is one more thing in the `_port_cmd_power_off` which can make the command fail. Not that it needs to be handled differently in your changes :slightly_smiling_face: 

> After it enables clock the the peripheral, it power-off the port, right?

And yes, this is correctly handled. First we must enable the peripheral clock, then power off the peripheral.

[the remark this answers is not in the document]
tore-espressif: Thank you for your review! I fixed the comment and also cleaned-up git history

host/usb/src/hub.c
```diff
@@ -620,10 +688,11 @@ static esp_err_t root_port_disable(void)
 
 esp_err_t hub_install(hub_config_t *hub_config, void **client_ret)
 {
+    HUB_DRIVER_CHECK(hub_config != NULL && client_ret != NULL && hub_config->port_map != 0, ESP_ERR_INVALID_ARG);
```
igi540: Nice addition validating port_map != 0. As a follow-up, it might be worth tightening validation a bit further: reject any bits beyond HCD_NUM_PORTS, and potentially fail early for combinations we don’t fully support yet (e.g. suspend/resume behavior in multi-port mode). That way we avoid silent “ignored bits” and make misconfiguration easier to diagnose.

igi540 (approved):

PR #420, opened 2026-02-19
fix(host/uac): Prevent integer overflow

tore-espressif:
fix(host/uac): Prevent integer overflow

### DIFF
--- a/host/class/uac/usb_host_uac/uac_host.c
+++ b/host/class/uac/usb_host_uac/uac_host.c
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2024-2025 Espressif Systems (Shanghai) CO LTD
+ * SPDX-FileCopyrightText: 2024-2026 Espressif Systems (Shanghai) CO LTD
  *
  * SPDX-License-Identifier: Apache-2.0
  */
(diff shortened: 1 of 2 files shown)


peter-marcisovsky (approved):
LGTM

host/class/uac/usb_host_uac/uac_host.c
```diff
@@ -2358,14 +2363,19 @@ esp_err_t uac_host_device_start(uac_host_device_handle_t uac_dev_handle, const u
     // enqueue multiple transfers to make sure the data is not lost
     iface->xfer_num = CONFIG_UAC_NUM_ISOC_URBS;
     iface->packet_num = CONFIG_UAC_NUM_PACKETS_PER_URB;
-    iface->packet_size = iface->iface_alt[iface->cur_alt].cur_sampling_freq * stream_config->channels * stream_config->bit_resolution / 8 / 1000;
+    // Use 64-bit intermediate to prevent overflow in packet size calculation
+    const uint64_t packet_size_calc = (uint64_t)iface->iface_alt[iface->cur_alt].cur_sampling_freq
+                                      * stream_config->channels
+                                      * stream_config->bit_resolution / 8;
+    iface->packet_size = (uint32_t)(packet_size_calc / 1000);
     iface->flags |= stream_config->flags;
```
igi540: Small concern: this updates `iface->flags` before the new packet size validation can fail. If the function returns `ESP_ERR_INVALID_SIZE` here, the interface stays in `IDLE`, but the requested flags remain latched. That can affect the next `uac_host_device_start()` call; for example, `FLAG_STREAM_SUSPEND_AFTER_START` may carry over and keep a later successful start suspended unexpectedly. Could we move this assignment until after all validation succeeds, or restore the previous flags on the error path?

igi540 (approved):

PR #424, opened 2026-02-23
fix(host/uvc): Allow high bandwidth endpoints

tore-espressif:
Closes https://github.com/espressif/esp-usb/issues/279

High-bandwidth endpoints are correctly selected and handled

image

### DIFF
--- a/host/class/uvc/usb_host_uvc/uvc_descriptor_parsing.c
+++ b/host/class/uvc/usb_host_uvc/uvc_descriptor_parsing.c
@@ -30,7 +30,14 @@ static inline uint16_t uvc_desc_get_safe_wTotalLength(const usb_config_desc_t *c
 {
     uintptr_t cfg_start = (uintptr_t)cfg_desc;
     uintptr_t header_start = (uintptr_t)header;
-    uint16_t remaining = cfg_desc->wTotalLength - (uint16_t)(header_start - cfg_start);
+    if (header_start <= cfg_start) {
+        return 0;
+    }
+    const size_t offset = (size_t)(header_start - cfg_start);
+    if (offset >= cfg_desc->wTotalLength) {
+        return 0;
+    }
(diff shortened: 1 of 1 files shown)


tore-espressif:
@peter-marcisovsky @igi540 This is small fix + test update for high-bandwidth endpoints that can offer up to 3 transactions per microframe.

In UVC cameras, this is typically needed when uncompressed (eg YUY2) formats are selected

host/class/uvc/usb_host_uvc/uvc_descriptor_parsing.c
```diff
@@ -30,7 +30,8 @@ static inline uint16_t uvc_desc_get_safe_wTotalLength(const usb_config_desc_t *c
 {
     uintptr_t cfg_start = (uintptr_t)cfg_desc;
     uintptr_t header_start = (uintptr_t)header;
-    uint16_t remaining = cfg_desc->wTotalLength - (uint16_t)(header_start - cfg_start);
+    assert(header_start > cfg_start);
+    size_t remaining = cfg_desc->wTotalLength - (uint16_t)(header_start - cfg_start);
```
igi540: this helper is intended to protect against malformed descriptors, so relying on `assert(...)` is not enough (asserts can be compiled out in release builds). Also, casting `(header_start - cfg_start)` to `uint16_t` can truncate if the header is outside config descriptor bounds.

Could we add a tiny runtime bounds check and compute `remaining` from a `size_t offset`?  
This keeps behavior identical for valid descriptors, but avoids underflow/truncation on invalid input.

```suggestion
    if (header_start < cfg_start) {
        return 0;
    }
    size_t offset = (size_t)(header_start - cfg_start);
    if (offset > cfg_desc->wTotalLength) {
        return 0;
    }
    size_t remaining = (size_t)cfg_desc->wTotalLength - offset;

```

PR #430, opened 2026-03-02
docs(host): Add note about max transfer size

tore-espressif:
Added a note about maximum transfer sizes.

We will check this in runtime after relevant changes in esp-idf/esp_hal_usb are merged.

## Related
- Original bug report https://github.com/espressif/esp-idf/issues/17927
- Original PR with fix https://github.com/espressif/esp-usb/pull/347

### DIFF
--- a/host/usb/include/usb/usb_host.h
+++ b/host/usb/include/usb/usb_host.h
@@ -662,6 +662,11 @@ esp_err_t usb_host_endpoint_clear(usb_device_handle_t dev_hdl, uint8_t bEndpoint
  * - A transfer object can be re-used indefinitely
  * - A transfer can be submitted using usb_host_transfer_submit() or usb_host_transfer_submit_control()
  *
+ * @note Maximum transfer size depends on hardware configuration and endpoint bMaxPacketSize, thus it is not checked on allocation.
+ *       Maximum transfer size is determined as follows:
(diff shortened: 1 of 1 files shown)


host/usb/include/usb/usb_host.h
```diff
@@ -662,6 +662,11 @@ esp_err_t usb_host_endpoint_clear(usb_device_handle_t dev_hdl, uint8_t bEndpoint
  * - A transfer object can be re-used indefinitely
  * - A transfer can be submitted using usb_host_transfer_submit() or usb_host_transfer_submit_control()
  *
+ * @note Maximum transfer size depends on hardware configuration and endpoint bMaxPacketSize, thus it is not checked on allocation.
+ *       Maximum transfer size is determined as follows:
+ *       - ESP32-S2, S3, H4 and P4 version < 3 = MIN(65535,   127 * bMaxPacketSize)
```
igi540: Nit/question: the note says *"ESP32-P4 version < 3 / >= 3"* — can we clarify what “version” refers to?

From the wording, I *think* this is the **chip revision** as configured via `menuconfig` (**Component config → Hardware Settings → Chip revision**, e.g. “Select ESP32-P4 revisions <3.0 / >=3.0”), but it’s not obvious to readers.

If that’s what you mean, could we rename it to **“ESP32-P4 chip revision”** (or “P4 revision”) for consistency with Kconfig/docs and avoid confusion with USB/IP/driver versions? If it’s something else, please point to the source of truth (e.g. which field/API defines it).

PR #426, opened 2026-02-24
Add configurable vendor RX and TX buffer sizes (IEC-493)

DeqingSun:
This allows user to use KConfig to change CFG_TUD_VENDOR_TX_BUFSIZE instead of edit the component file. It helps to remove ZLP (ADB emulation) or send a large chunk of data.

## Description

I was trying to implement the ADB device on ESP32P4 and the ZLP became an issue on MacOS. An easy way to bypass ZLP is to set CFG_TUD_VENDOR_TX_BUFSIZE to 0. However, I can not override the ```device/esp_tinyusb/include/tusb_config.h```. The proposed change allows change of  CFG_TUD_VENDOR_TX_BUFSIZE with sdkconfig, and keeps the tinyusb component untouched, easier for archiving code.

## Testing

When I add ```config CFG_TUD_VENDOR_TX_BUFSIZE``` in Kconfig.projbuild, all CFG_TUD_VENDOR_TX_BUFSIZE will be changed.

---

## Checklist

Before submitting a Pull Request, please ensure the following:

- [ ] 🚨 This PR does not introduce breaking changes.
- [ ] All CI checks (GH Actions) pass.
- [ ] Documentation is updated as needed.
- [ ] Tests are updated or added as necessary.
- [ ] Code is well-commented, especially in complex areas.
- [ ] Git history is clean — commits are squashed to the minimum necessary.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes only add optional Kconfig-backed overrides for vendor class buffer sizes, with existing defaults preserved unless enabled. Main risk is misconfiguration leading to increased RAM use or unexpected USB transfer behavior (e.g., ZLP handling).
> 
> **Overview**
> Adds Kconfig options to optionally override TinyUSB Vendor Class RX/TX FIFO sizes when `TINYUSB_VENDOR_COUNT > 0`, exposing `TINYUSB_CFG_TUD_VENDOR_RX_BUFSIZE` and `TINYUSB_CFG_TUD_VENDOR_TX_BUFSIZE` behind enable toggles.
> 
> Updates `tusb_config.h` so `CFG_TUD_VENDOR_RX_BUFSIZE`/`CFG_TUD_VENDOR_TX_BUFSIZE` use these `sdkconfig` values when set, otherwise retaining the existing speed-based defaults (512 HS / 64 FS).
> 
[...]

PR #432, opened 2026-03-02
feat(esp_tinyusb): Add Kconfig option for Vendor specific buffers

tore-espressif:
Add option to configure Vendor specific class buffer via menuconfig

Closes https://github.com/espressif/esp-usb/pull/426

### DIFF
--- a/device/esp_tinyusb/Kconfig
+++ b/device/esp_tinyusb/Kconfig
@@ -318,5 +318,41 @@ menu "TinyUSB Stack"
             range 0 2
             help
                 Setting value greater than 0 will enable TinyUSB Vendor specific feature.
+
+        config TINYUSB_VENDOR_RX_BUFSIZE
+            depends on (TINYUSB_VENDOR_COUNT > 0)
+            int "Vendor specific FIFO size of RX channel"
+            default 512 if IDF_TARGET_ESP32P4
+            default 64
+            range 0 32768
+            help
+                This buffer size defines maximum data length in bytes that you can receive at once.
+
+                Note: Unlike other USB classes, the vendor class supports setting buffer sizes to 0
+                to disable internal buffering. When disabled, RX data goes directly to tud_vendor_rx_cb(),
+                the tud_vendor_read() function is not available and application must handle data in the callback.
+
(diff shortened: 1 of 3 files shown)


device/esp_tinyusb/Kconfig
```diff
@@ -318,5 +318,33 @@ menu "TinyUSB Stack"
...
             help
                 Setting value greater than 0 will enable TinyUSB Vendor specific feature.
+
+        config TINYUSB_VENDOR_RX_BUFSIZE
+            depends on (TINYUSB_VENDOR_COUNT > 0)
+            int "Vendor specific FIFO size of RX channel"
+            default 512 if IDF_TARGET_ESP32P4
+            default 64
+            range 64 10000
+            help
```
peter-marcisovsky: There is a comment in tinyusb regarding both, the RX and TX buffers, which might be worth mentioning in help:

>// RX FIFO can be disabled by setting this value to 0
#ifndef CFG_TUD_VENDOR_RX_BUFSIZE
#define CFG_TUD_VENDOR_RX_BUFSIZE    64
#endif

device/esp_tinyusb/Kconfig
```diff
@@ -318,5 +318,39 @@ menu "TinyUSB Stack"
...
+            depends on (TINYUSB_VENDOR_COUNT > 0)
+            int "Vendor specific FIFO size of TX channel"
+            default 512 if IDF_TARGET_ESP32P4
+            default 64
+            help
+                This buffer size defines maximum data length in bytes that you can transmit at once.
+
+                Note: Unlike other USB classes, the vendor class supports setting buffer sizes to 0
+                to disable internal buffering. When disabled, tud_vendor_write() bypasses internal software FIFO,
+                and the passed buffer is passed to USB DMA.
```
igi540: Minor wording suggestion: when `TINYUSB_VENDOR_TX_BUFSIZE` is `0`, `tud_vendor_write()` does bypass the software FIFO, but it still copies into TinyUSB's internal endpoint buffer before starting the transfer. Because of that, saying the application buffer is "passed to USB DMA" seems a bit misleading.

igi540 (approved):

PR #445, opened 2026-03-20
feat(usb_host): esp sleep modes verification

peter-marcisovsky:
## Description

Verification of using light sleep and deep sleep in combination with USB Host application.

## Light sleep

- Using `CONFIG_ESP_SLEEP_EVENT_CALLBACKS`
- Register enter and exit light sleep events
- Calling root port suspend (not public API, but a "faster" HCD command) upon enter light sleep to effectively suspend the peripheral and the connected device
- Connected device enters suspended state during the light sleep
- Connected device must not undergo a reset during light sleep

#### Light sleep log, with light sleep callbacks enabeld

Root port enters suspended mode before entering light sleep

Screenshot from 2026-03-27 12-31-41

#### Light sleep log, with light sleep callbacks disabled

Missing the suspend sequence from the host. Host just stops sending SOFs without any "reason". This pattern only appears on HS port. The device does not enter suspended state

Screenshot from 2026-04-01 15-42-00

### Enter light sleep latency

Added light sleep latency when entering and exiting light sleep with automatic suspend of the peripheral

|                                  | esp32s2 | esp32s3 | esp32p4 | esp32h4 | esp32s31 |
|----------------------------|-------------|-------------|---------------|--------------|---------------|
| enter light sleep    | ~140 uS |  ~90 uS   | ~13 uS     | ~60 uS    | ~35 uS     |
| exit light sleep       | ~55 uS   |  ~40 uS    | ~6 uS      |  ~35 uS    | ~30 uS     |

### Enter light sleep callback
- Auto suspends the root port in (if not already suspended)
- Synchronously halts and flushes all the EPs of all devices
- Using fast suspend sequence, suspends the root port
  - does not wait until the bus settles (20 mS) thus expects that the root port finished suspend event successfully

### Exit light sleep callback
- Delivers suspended client event (if the enter light sleep callback suspended the root port)
- Checks if the root port did not undergo disconnect or error state
  - If yes, no suspend event is delivered to clients

## Deep sleep

- Connected device shall be disconnected upon entering deep sleep and connected back after exiting deep sleep
- RAM is unpowered during deep sleep and any information about the connected device (cached descriptors, device state.. ) is lost during deep sleep
- Currently, `HNP` cap is enabled on all `HNP` capable targets by default, to prevent current leakage in deep sleep.

## Testing

- `usb_sleep_modes` target tests
  - light sleep functionality verification
  - deep sleep functionality verification
  - enter/exit light sleep latency
  - error handling inside the enter/exit light sleep callbacks
  - client events delivery

## Limitations

- Light sleep: a device must be present in order to call root port suspend in a light sleep callback
  - Known "issue" from the original suspend/resume implementation before HCLK clock gating was introduced
  - Will be fixed in a separate PR, which will remove the device presence obligation when entering light sleep

## Related
- Documentation #476 
- Tested by #478 
---

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds sleep event callbacks that suspend/resume the USB root port from light-sleep critical sections and introduces new synchronization/error-handling paths, which could affect USB stability and wake latency on supported targets.
> 
> **Overview**
> Adds optional **automatic USB root-port suspend/resume around light sleep** (guarded by `CONFIG_ESP_SLEEP_EVENT_CALLBACKS`) by registering sleep event callbacks in `usb_host_install()`/`usb_host_uninstall()` and coordinating async suspend completion via a new semaphore.
> 
> Extends the hub driver with a `HUB_EVENT_SUSPEND_COMPLETED` event and a `hub_root_mark_light_sleep_auto_resume()` flag to safely handle the case where a device disconnects during light sleep (HCD port in `RECOVERY`) before attempting resume.
> 
> Introduces a new `usb_sleep_modes` target test app (Unity + pytest) that exercises USB Host behavior across repeated light-sleep cycles and multi-stage deep-sleep resets, including an error-path test for sleep callback failures.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7a8ee7ad7f56076367739f343649f35e56b88b01. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


host/usb/src/hub.c
```diff
@@ -593,9 +605,26 @@ static void root_port_req(root_hub_port_t *root_hub_port)
...
+        const hcd_port_state_t hcd_port_state = hcd_port_get_state(root_port_hdl);
+        if (hcd_port_state == HCD_PORT_STATE_RECOVERY) {
+            if (light_sleep_auto_resume) {
+                // We are resuming root port from light sleep and the HCD port is currently in recovery state
+                // That indicates, that the device has been disconnected during light sleep
+                // Clock gating will be disabled by recovery routine, nothing to do here
+                return;
+            }
+            // root port in unexpected (recovery) state, fatal error, do ESP_ERROR_CHECK for backtrace
+            ESP_ERROR_CHECK(hcd_port_state == HCD_PORT_STATE_RECOVERY);
```
tore-espressif: Will this always be true given the `if` on line 616?

host/usb/src/usb_host.c
```diff
@@ -525,6 +540,109 @@ static void auto_suspend_timer_cb(TimerHandle_t xTimer)
...
+ * @return
+ *    - ESP_OK: Root port successfully resumed, normal operation restored after light sleep
+ *    - Other error codes from usb_host_lib_root_port_resume()
+ */
+static esp_err_t exit_light_sleep(void *user_arg, void *ext_arg)
+{
+    ESP_EARLY_LOGD(USB_HOST_TAG, "Exit light sleep cb");
+    // Using ISR variant, because we are in a light sleep critical section
+    ESP_RETURN_ON_ERROR_ISR(hub_root_mark_light_sleep_auto_resume(), USB_HOST_TAG, "Root port in incorrect state");
+    ESP_RETURN_ON_ERROR_ISR(usb_host_lib_root_port_resume(), USB_HOST_TAG, "Failed to resume root port after light sleep");
```
peter-marcisovsky, line 620: One potential concurrency issue here will be fixed in a follow-up:

`usb_host_lib_root_port_resume()` -> `usbh_devs_num()` -> `xSemaphoreTake(p_usbh_obj->constant.mux_lock, portMAX_DELAY);`

The `usbh_devs_num` takes a `portMAX_DELAY` mutex, which can deadlock, considering this callback is called from a critical section of light sleep. 

Proposed fixes:

- fully deferred `usb_host_lib_root_port_resume` call using `usb_host_lib_process`
- or remove the `usbh_devs_num()` completely and getting the HCD port state instead as an indicator whether the device is connected or not, since we want to call the suspend/resume even with no devices connected

Same comment for the suspend API and the light sleep enter callback

peter-marcisovsky:
Converted to draft, trying to lower the sleep latency even more.

host/usb/src/hub.c
```diff
@@ -593,9 +605,28 @@ static void root_port_req(root_hub_port_t *root_hub_port)
...
```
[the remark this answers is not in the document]
peter-marcisovsky: ESP_ERROR_CHECK is present only for a backtrace. I asserts when the root port is in an unexpected recovery state (not caused by the disconnection in the light sleep).

Not relevant.

host/usb/test/target_test/usb_sleep_modes/main/test_usb_host_esp_sleep.c
```diff
@@ -0,0 +1,326 @@
...
+{
+    EventGroupHandle_t test_event_group = (EventGroupHandle_t) args;
+    // Configure timer wakeup source
+    TEST_ASSERT_EQUAL(ESP_OK, esp_sleep_enable_timer_wakeup(TIMER_DEEP_SLEEPWAKEUP_TIME_US));
+    ESP_LOGI(USB_SLEEP_TAG, "Deep sleep timer wakeup source is ready");
+
+    TEST_ASSERT_EQUAL_MESSAGE(
+        pdTRUE,
+        xEventGroupWaitBits(test_event_group, EVT_ALLOW_SLEEP | EVT_CLIENT_CLOSE, pdTRUE, pdFALSE, pdMS_TO_TICKS(EVT_WAIT_TIME_MS)),
+        "CTRL client did not set event on time to allow deep sleep");
```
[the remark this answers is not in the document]
peter-marcisovsky: Fixed

host/usb/src/usbh.c
```diff
@@ -1064,6 +1066,70 @@ esp_err_t usbh_devs_mark_all_free(void)
...
+    for (int i = 0; i < NUM_NON_DEFAULT_EP; i++) {
+        if (dev_obj->mux_protected.endpoints[i] != NULL) {
+            ESP_ERROR_CHECK(hcd_pipe_command(dev_obj->mux_protected.endpoints[i]->constant.pipe_hdl, HCD_PIPE_CMD_HALT));
+            ESP_ERROR_CHECK(hcd_pipe_command(dev_obj->mux_protected.endpoints[i]->constant.pipe_hdl, HCD_PIPE_CMD_FLUSH));
+        }
+    }
+    xSemaphoreGive(p_usbh_obj->constant.mux_lock);
+
+    ESP_ERROR_CHECK(hcd_pipe_command(dev_obj->constant.default_pipe, HCD_PIPE_CMD_HALT));
+    ESP_ERROR_CHECK(hcd_pipe_command(dev_obj->constant.default_pipe, HCD_PIPE_CMD_FLUSH));
```
tore-espressif: can we just do:
```suggestion
    handle_epn_halt_flush(dev_obj);
    handle_ep0_flush(dev_obj);
```

to reduce copy-pasted code?

host/usb/src/usbh.c
```diff
@@ -1064,6 +1066,70 @@ esp_err_t usbh_devs_mark_all_free(void)
...
+ *
+ * @return
+ *    - ESP_OK: Halted and Flushed successfully
+ */
+static esp_err_t halt_flush_one_device(device_t *dev_obj)
+{
+    xSemaphoreTake(p_usbh_obj->constant.mux_lock, portMAX_DELAY);
+    for (int i = 0; i < NUM_NON_DEFAULT_EP; i++) {
+        if (dev_obj->mux_protected.endpoints[i] != NULL) {
+            ESP_ERROR_CHECK(hcd_pipe_command(dev_obj->mux_protected.endpoints[i]->constant.pipe_hdl, HCD_PIPE_CMD_HALT));
```
tore-espressif: In case this call fails when called from light sleep enter callback, the whole system will abort.

We can keep it as it is for now (I can see that we use the same pattern in `handle_epn_halt_flush()` already), but we should keep this in mind

host/usb/src/hub.c
```diff
@@ -884,6 +914,45 @@ esp_err_t hub_root_stop(void)
...
+        root_hub_port = &p_hub_driver_obj->root_hub_ports[1];
+    }
+#endif
+    assert(root_hub_port->constant.hdl);
+    HUB_DRIVER_CHECK_FROM_CRIT(root_hub_port->dynamic.state == ROOT_PORT_STATE_ENABLED, ESP_ERR_NOT_ALLOWED);
+    hcd_port_handle_t root_port_hdl = root_hub_port->constant.hdl;
+    HUB_DRIVER_EXIT_CRITICAL();
+
+    ESP_RETURN_ON_ERROR(usbh_devs_halt_flush_all_sync(), HUB_DRIVER_TAG, "Devs halt/flush failed");
+    ESP_RETURN_ON_ERROR(hcd_port_command(root_port_hdl, HCD_PORT_CMD_SUSPEND_LIGHT_SLEEP), HUB_DRIVER_TAG, "Port suspend cmd failed");
```
tore-espressif: Should we call `hub_root_can_suspend` between these two calls? Or is this intentionally omitted?

host/usb/src/usbh.c
```diff
@@ -1064,6 +1066,70 @@ esp_err_t usbh_devs_mark_all_free(void)
...
+        }
+    }
+    xSemaphoreGive(p_usbh_obj->constant.mux_lock);
+
+    ESP_ERROR_CHECK(hcd_pipe_command(dev_obj->constant.default_pipe, HCD_PIPE_CMD_HALT));
+    ESP_ERROR_CHECK(hcd_pipe_command(dev_obj->constant.default_pipe, HCD_PIPE_CMD_FLUSH));
+    return ESP_OK;
+}
+
+esp_err_t usbh_devs_halt_flush_all_sync(void)
```
tore-espressif, line 1119: 1. Why do we want to avoid taking the mutex in this function?
2. Could we pass a 'halt+flush all pipes' command to HCD? Eliminating the walking of USBH device list altogether?

host/usb/test/target_test/usb_sleep_modes/main/test_usb_host_esp_sleep.c
```diff
@@ -0,0 +1,537 @@
...
+    return ESP_OK;
+}
+
+/**
+ * @brief Light sleep callback registered before installing usb_host_lib, invoked when entering light sleep (suspending)
+ */
+static const esp_sleep_event_cb_config_t s_before_auto_suspend_cb = {
+    .cb = test_time_before_auto_suspend,
+    .user_arg = (void *) &s_light_sleep_time_stamp,
+    .prior = 2,
```
tore-espressif, line 422: How do you guarantee that the callback is called before or after when the priority is always = 2?

host/usb/test/target_test/usb_sleep_modes/main/test_usb_host_esp_sleep.c
```diff
@@ -0,0 +1,537 @@
...
+    .next = NULL,
+};
+
+/**
+ * @brief Light sleep callback registered after installing usb_host_lib, invoked when entering light sleep (suspending)
+ */
+static const esp_sleep_event_cb_config_t s_after_auto_suspend_cb = {
+    .cb = test_time_after_auto_suspend,
+    .user_arg = (void *) &s_light_sleep_time_stamp,
```
tore-espressif, line 432: Nitpick, does not have to be changed:
instead of passing `s_light_sleep_time_stamp` as user argument, we could pass `s_light_sleep_time_stamp.after_suspend`.
This way there would be only one `test_time_xxx()` callback

```suggestion
    .user_arg = (void *) &(s_light_sleep_time_stamp.after_suspend),
```

host/usb/src/hub.c
```diff
@@ -623,8 +644,17 @@ static void root_port_req(root_hub_port_t *root_hub_port)
...
+        if (light_sleep_auto_resume) {
+            // Resuming after light sleep
+            // Deliver deferred client suspend notifications first (usbh_process order), then resume EPs and deliver resume evt
+            usbh_devs_set_pm_actions_all(USBH_DEV_SUSPEND_EVT | USBH_DEV_RESUME | USBH_DEV_RESUME_EVT);    // NOLINT(clang-analyzer-optin.core.EnumCastOutOfRange)
+        } else
+#endif // AUTO_PM_LIGHT_SLEEP
+        {
+            // Clear all EPs and propagate the resumed event to clients
+            usbh_devs_set_pm_actions_all(USBH_DEV_RESUME | USBH_DEV_RESUME_EVT);    // NOLINT(clang-analyzer-optin.core.EnumCastOutOfRange)
+        }
```
tore-espressif, line 651: Would this be more readable?
```suggestion
        uint32_t actions = USBH_DEV_RESUME | USBH_DEV_RESUME_EVT;
#ifdef AUTO_PM_LIGHT_SLEEP
        if (light_sleep_auto_resume) {
            actions |= USBH_DEV_SUSPEND_EVT;
        }
#endif // AUTO_PM_LIGHT_SLEEP
        // Clear all EPs and propagate the resumed event to clients
        usbh_devs_set_pm_actions_all(actions)
```

host/usb/src/usb_host.c
```diff
@@ -525,6 +530,82 @@ static void auto_suspend_timer_cb(TimerHandle_t xTimer)
     HOST_EXIT_CRITICAL();
 }
 
+#ifdef AUTO_PM_LIGHT_SLEEP
+/**
+ * @brief Callback function invoked before entering light sleep
+ *
+ * - Synchronously halt/flush all endpoints, assert DWC root port suspend without HCD internal clock gating, and mark
```
tore-espressif:
```suggestion
 * - Synchronously halt/flush all endpoints, set DWC root port suspend bit without HCD internal clock gating, and mark
```

host/usb/src/usb_host.c
```diff
@@ -525,6 +530,82 @@ static void auto_suspend_timer_cb(TimerHandle_t xTimer)
...
+ * @param[in] ext_arg External argument, not used
+ * @return
+ *    - ESP_OK: Root port successfully resumed, normal operation restored after light sleep
+ *    - ESP_ERR_INVALID_STATE: usb host lib is not installed
+ *    - Other error codes from hub_root_mark_resume()
+ */
+static esp_err_t exit_light_sleep(void *user_arg, void *ext_arg)
+{
+    HOST_CHECK(p_host_lib_obj != NULL, ESP_ERR_INVALID_STATE);
+    ESP_EARLY_LOGD(USB_HOST_TAG, "Exit light sleep cb");
```
tore-espressif: Question: Do we _always_ want to resume after we exit light sleep?

Eg. if the port was suspended when we entered light sleep, we should not resume, right?

tore-espressif (commented):
@peter-marcisovsky Very nice work!
thank you for the tests and paying attention to light sleep enter/exit latencies.

Overall the approach LGTM. I left few comments and a few nitpicks

host/usb/src/usb_host.c
peter-marcisovsky: This is a good point and is't related to a follow-up PR we discussed. In which we will add a Kconfig, or some configuration struct for the user to decide whether he wants to automatically resume root port when exiting light sleep (same for the auto suspend when entering light sleep).

For example in a case, where the a user would be oscilating between light sleep and normal operation too fast that it would become unable for the USB Host to complete the resume procedure (which is something like 20ms AFAIK) and to deliver resume events to clients.

host/usb/src/usbh.c
peter-marcisovsky: I had the same dilema when working on this. But `hcd_pipe_command` only recycles return values for the commands it's executing and the  `HCD_PIPE_CMD_HALT` command returns only `ESP_OK`.

Exactly, we are using the same pattern across the stack.

host/usb/src/usbh.c
peter-marcisovsky: I added `usbh_devs_halt_flush_all_sync()`, because normally we:
- set device actions (EP Halt, EP flush)
- we wait for the `usb_host_lib_handle_events` to finish the device actions

Now, being in the light sleep callback we would have to:
- set the device actions
- wait for the `usb_host_lib_handle_events` to finish all the actions (maybe even some already pending ones) which would take time.
- create some event signalizing that all EPs of all devices are halted and flushed, probably with semaphore
- enter light sleep

To decrease the enter sleep latency I added this synchronous approach which does not depend on the `usb_host_lib_handle_events` and on current load of the usb_host_lib. 

I can try doing it the usual way. Just set the actions and wait for the completion for comparison.

host/usb/src/usbh.c
peter-marcisovsky: Can do, I must have missed this or trying something else and forgot to update it. Will fix it.

host/usb/test/target_test/usb_sleep_modes/main/test_usb_host_esp_sleep.c
peter-marcisovsky: Either priority, or order of callbacks registration depends, explanation [here](https://github.com/espressif/esp-idf/blob/master/components/esp_hw_support/include/esp_private/sleep_event.h#L61-L66)

I firstly tired different priorities, but that kinda did not work for some reason. Now, the callbacks are executed in the same order they were registered.

tore-espressif: Thanks for the explanation!

host/usb/src/hub.c
peter-marcisovsky: This is outdated as the code has changed.

host/usb/src/hub.c
peter-marcisovsky: Now the `hub_root_light_sleep_suspend_bus()` checks the root hub state and the hcd port state (same as the `hub_root_can_suspend()` ) checks and decides whether it's possible to continue with the auto suspend or not.

host/usb/src/usbh.c
peter-marcisovsky: I modified the `usbh_devs_halt_flush_all_sync` function so it now acquires a mutex once in the beginning and processess all the EPs of all the device withn that mutex so the mutex is not acquired/released per device but only once per this function.

host/usb/test/target_test/usb_sleep_modes/main/test_usb_host_esp_sleep.c
peter-marcisovsky: I will keep the current implementation if that's ok.

host/usb/src/hub.c
peter-marcisovsky: This is not relevant anymore, as the code has changed.

host/usb/src/usb_host.c
peter-marcisovsky: Removed the auto resume when exiting from the light sleep.

peter-marcisovsky:
@tore-espressif  Thank you for the review. I removed the automatic resume when exiting light sleep. Now, only the auto suspend when entering light sleep in in place.
I kept exit light sleep callback thought (used as auto resume before) to:
- deliver the suspend event to the clients after light sleep exit
- handle possible device disconnection during the light sleep (where suspend event will not be delivered)

I remeasured the enter/exit light sleep latency and it has improved as both callbacks are now simpler (fewer mutexes and fewer critical sections).

Recent changes in separate commit. I have some more ideas for improvement, but I would keep it for separate PRs, as this one is already too big.

host/usb/src/usbh.c
```diff
@@ -1063,6 +1114,41 @@ esp_err_t usbh_devs_mark_all_free(void)
     return (wait_for_free) ? ESP_ERR_NOT_FINISHED : ESP_OK;
 }
 
+#ifdef AUTO_PM_LIGHT_SLEEP
+
+esp_err_t usbh_devs_halt_flush_all_sync(void)
+{
+    esp_err_t ret;
+    USBH_CHECK(p_usbh_obj != NULL, ESP_ERR_INVALID_STATE);
+    xSemaphoreTake(p_usbh_obj->constant.mux_lock, portMAX_DELAY);
```
peter-marcisovsky: This is one of the concerns about the whole feature. Having a blocking calls inside the callbacks. I found [this](https://github.com/espressif/esp-idf/blob/master/components/esp_hw_support/include/esp_private/sleep_event.h#L94) comment in the docs, about the blocking calls.

In our case, we are using `SLEEP_EVENT_SW_GOTO_SLEEP` and `SLEEP_EVENT_SW_EXIT_SLEEP`, which are SW events, not HW events. Thus using the blocking calls might not be strictly related to the our case, as the docs say:
> some of these callback functions

One of the possible (or partial) solutions could be just try to acquire the mutex without blocking. If mutex is not acquired, we just bail on auto suspend and prioritize light sleep latency.

tore-espressif: @peter-marcisovsky for this PR, I'd recommend
``` c
if (xSemaphoreTake(p_usbh_obj->constant.mux_lock, 0) != pdTRUE) {
    return ESP_ERR_NOT_FINISHED;
```

Have you got a condition where this semaphoretake call would actually block? the scheduler should be suspend when we call this, so it should assert...
In follow-ups we can improve this.

tore-espressif:
@peter-marcisovsky 

1. The feature LGTM. I'll do more testing on real HW and then approve.
2. Since we plan to have follow-ups for this feature, should we release https://github.com/espressif/esp-usb/milestone/38 first, so it is not blocked by system_sleep PRs?

host/usb/src/usbh.c
peter-marcisovsky: I have not hit any conditions yet. One of those could be calling suspend from another task, or connect/disconnect, enumeration.. being asynchronous events. 
I did the change as suggested. I will consider adding some tests which would prove this behavior in a follow-up.

host/usb/src/usb_host.c
```diff
@@ -525,6 +530,83 @@ static void auto_suspend_timer_cb(TimerHandle_t xTimer)
...
+ *    - Other errors from hub_root_light_sleep_suspend_bus()
+ */
+static esp_err_t enter_light_sleep(void *user_arg, void *ext_arg)
+{
+    ESP_LOGV(USB_HOST_TAG, "Enter light sleep cb");
+    HOST_ENTER_CRITICAL();
+    HOST_CHECK_FROM_CRIT(p_host_lib_obj != NULL, ESP_ERR_INVALID_STATE);
+    HOST_EXIT_CRITICAL();
+
+    ESP_RETURN_ON_ERROR(hub_root_light_sleep_suspend_bus(), USB_HOST_TAG, "Failed to suspend root port before light sleep");
```
tore-espressif, line 555: This will print an error, in case we go to light sleep and the port is Not_powered or Powered.

In this case, we do not want to suspend the port so the log is a little confusing...

tore-espressif (approved):
Overall LGTM!

host/usb/src/usb_host.c
peter-marcisovsky: Thanks for the testing, will fix the log.

peter-marcisovsky: Removed the error when trying to auto-suspend the port, but the port is not powered off or no device is connected.

PR #467, opened 2026-04-10
feat(esp_tinyusb): Support ESP32-S31 as HS-only single-port device

igi540:
## Description
This PR enables `esp_tinyusb` on ESP32-S31, which has a single HS-only USB-OTG peripheral (UTMI PHY, no internal FSLS PHY).
Previously, HS handling in `esp_tinyusb` assumed that HS implies a multi-port chip (checked via `SOC_USB_OTG_PERIPH_NUM > 1`, with the HS port being `TINYUSB_PORT_HIGH_SPEED_0`). ESP32-S31 breaks that assumption: it has only one port, and that port is always HS. The refactor replaces those checks with two small helpers:
- `TINYUSB_PORT_IS_HS(port)` — used during descriptor selection in `descriptors_control.c`
- `TINYUSB_PORT_DEFAULT_HS` — used as the default port value in `TINYUSB_CONFIG_HIGH_SPEED()`
so both multi-port HS chips (e.g. ESP32-P4) and single-port HS-only chips (e.g. ESP32-S31) are handled uniformly.
Other notable changes:
- TinyUSB dependency is bumped to `>=0.19.0~3-rc1`, which is required for ESP32-S31 support.
- The DCache enable condition in `tusb_config.h` is tightened: having an L1 cache (`CONFIG_CACHE_L1_CACHE_LINE_SIZE != 0`) is no longer sufficient — internal SRAM must also be routed through it (`CONFIG_SOC_CACHE_INTERNAL_MEM_VIA_L1CACHE`). Otherwise DCache clean/invalidate calls on TinyUSB DMA buffers (which live in internal SRAM) are unnecessary and incorrect on S31.
- Software VBUS monitoring (previously P4-only) is now also compiled in for ESP32-S31 and initialized unconditionally for self-powered devices on that target.
## Related
IDF-14703
## Testing
Manually verified on hardware:
- **ESP32-S31 (new target)**
  - `msc device example` — MSC device enumerates and works at high speed.
  - `hid device example` — HID device enumerates and works at high speed.
- **ESP32-S2 (regression check)**
  - `msc device example` — MSC device still enumerates and works at full speed, confirming the HS refactor does not break existing FS-only single-port targets.

---

> [!NOTE]
> **Medium Risk**
> Extends core USB init/descriptor and DMA-cache configuration logic to handle an HS-only target and bumps the TinyUSB dependency, which could affect enumeration and DMA behavior across supported chips if assumptions are wrong.
> 
> **Overview**
> Adds **ESP32-S31** as a supported target for `esp_tinyusb`, updating default config macros and descriptor selection to treat the single port as *always high-speed* (vs. the previous “HS implies multi-port” assumption).
> 
> Updates runtime init to use the UTMI PHY and HS speed on S31, enables software VBUS monitoring for S31 (and includes the GPIO dependency), and adjusts TinyUSB task port-speed selection accordingly.
> 
> Bumps the `tinyusb` component requirement to `>=0.19.0~3-rc1`, tightens the DMA DCache enable condition to require `CONFIG_SOC_CACHE_INTERNAL_MEM_VIA_L1CACHE`, and updates Kconfig buffer-size defaults/ranges for HS-capable targets.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5eb9b4db804b3b5beb77dc1868fe76a94e3dc1ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/device/esp_tinyusb/idf_component.yml
+++ b/device/esp_tinyusb/idf_component.yml
@@ -11,8 +11,12 @@ targets:
   - esp32s3
   - esp32p4
   - esp32h4
+  - esp32s31
 dependencies:
   idf: '>=5.0' # IDF 4.x contains TinyUSB as submodule
   tinyusb:
-    version: '>=0.17.0~2' # 0.17.0~2 is the first version that supports deinit
+    version: '>=0.19.0~3-rc1'
     public: true
+files:
+  exclude:
+    - "test_apps"
--- a/device/esp_tinyusb/idf_component.yml
+++ b/device/esp_tinyusb/idf_component.yml
@@ -11,8 +11,9 @@ targets:
   - esp32s3
   - esp32p4
   - esp32h4
+  - esp32s31
 dependencies:
   idf: '>=5.0' # IDF 4.x contains TinyUSB as submodule
   tinyusb:
-    version: '>=0.17.0~2' # 0.17.0~2 is the first version that supports deinit
+    version: '>=0.19.0~3-rc1'
     public: true
--- a/device/esp_tinyusb/tinyusb.c
+++ b/device/esp_tinyusb/tinyusb.c
@@ -171,13 +171,17 @@ esp_err_t tinyusb_driver_install(const tinyusb_config_t *config)
             .otg_speed = USB_PHY_SPEED_FULL,
         };
 
-#if (SOC_USB_OTG_PERIPH_NUM > 1)
+#if CONFIG_IDF_TARGET_ESP32S31
+        // ESP32-S31 has only UTMI (HS) PHY, no internal FSLS PHY
+        phy_conf.target = USB_PHY_TARGET_UTMI;
+        phy_conf.otg_speed = USB_PHY_SPEED_HIGH;
+#elif (SOC_USB_OTG_PERIPH_NUM > 1)
         if (config->port == TINYUSB_PORT_HIGH_SPEED_0) {
             // Default PHY for OTG2.0 is UTMI
             phy_conf.target = USB_PHY_TARGET_UTMI;
             phy_conf.otg_speed = USB_PHY_SPEED_HIGH;
         }
-#endif // (SOC_USB_OTG_PERIPH_NUM > 1)
+#endif // CONFIG_IDF_TARGET_ESP32S31
 
         // OTG IOs config
         const usb_phy_otg_io_conf_t otg_io_conf = USB_PHY_SELF_POWERED_DEVICE(config->phy.vbus_monitor_io);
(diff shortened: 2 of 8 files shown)


tore-espressif:
@igi540 I tested with tinyusb 0.19.0~3-rc1 and all seems to work correctly.

I did not test VBUS yet

device/esp_tinyusb/idf_component.yml
peter-marcisovsky, line 1: You can also exclude test app folder from being uploaded.

device/esp_tinyusb/tinyusb.c
```diff
@@ -171,13 +171,17 @@ esp_err_t tinyusb_driver_install(const tinyusb_config_t *config)
...
+        phy_conf.target = USB_PHY_TARGET_UTMI;
+        phy_conf.otg_speed = USB_PHY_SPEED_HIGH;
+#elif (SOC_USB_OTG_PERIPH_NUM > 1)
         if (config->port == TINYUSB_PORT_HIGH_SPEED_0) {
             // Default PHY for OTG2.0 is UTMI
             phy_conf.target = USB_PHY_TARGET_UTMI;
             phy_conf.otg_speed = USB_PHY_SPEED_HIGH;
         }
-#endif // (SOC_USB_OTG_PERIPH_NUM > 1)
+#endif // CONFIG_IDF_TARGET_ESP32S31
```
peter-marcisovsky, line 184:
```suggestion
#endif // CONFIG_IDF_TARGET_ESP32S31 || SOC_USB_OTG_PERIPH_NUM > 1
```

peter-marcisovsky (commented):
BTW, no esp32s31 apps are built in this CI. Both builds, the device test apps and esp-idf exaples are disabled for esp32s31.

tore-espressif:
> BTW, no esp32s31 apps are built in this CI. Both builds, the device test apps and esp-idf exaples are disabled for esp32s31.

@igi540 @peter-marcisovsky So we merge https://github.com/espressif/esp-usb/pull/469 now and then rebase this?

device/esp_tinyusb/tinyusb.c
igi540: The `#endif` here closes the entire `#if` / `#elif` chain, not a combined `||` condition. The preprocessor structure is:

```c
#if CONFIG_IDF_TARGET_ESP32S31
    ...
#elif (SOC_USB_OTG_PERIPH_NUM > 1)
    ...
#endif
```

These are **mutually exclusive** branches (`#if` / `#elif`), so the logic is "if ESP32-S31, do X; otherwise if multi-OTG, do Y." Changing the comment to `CONFIG_IDF_TARGET_ESP32S31 || SOC_USB_OTG_PERIPH_NUM > 1` would imply an `#if (A || B)` condition, which is a different logical construct — it would suggest both conditions lead to the **same** code block, which is not the case here.

The common convention is to annotate `#endif` with the opening `#if` condition, so `// CONFIG_IDF_TARGET_ESP32S31` is correct and consistent with standard practice. No change needed here.

peter-marcisovsky (approved):

tore-espressif (approved):
@igi540 @peter-marcisovsky I could not reproduce the example build failure locally... If all other tests pass, IMO we can merge this to unblock CI for other PRs and double check everthing during esp_tinyusb v2.2.0 release

peter-marcisovsky:
I have the same "error" locally, even thought it's just and warning during build, but warnings are treated as errors in CI `build succeeded with warnings`

```sh
(cmake) App examples/network/sta2eth, target esp32s2, sdkconfig (default), build in examples/network/sta2eth/build_esp32s2_default, build failed in 83.471661s: build succeeded with warnings
```
The app built fine. I suppose we can continue and catch this problem during `esp_tinyusb` release. I will take a look into it in meanwhile.

Issue #470, opened 2026-04-14
hid_host: disconnect cleanup aborts on endpoint_halt failure — USB context unrecoverable without physical re-plug (IEC-520)


munggo:
### Environment
- `espressif/usb_host_hid` v1.1.0 (same behavior on v1.2.0)
- ESP32-S3-WROOM-1-N16R8, ESP-IDF v5.2.5
- USB device: AliExpress **IOTC 0370** barcode scanner module (Waveshare-compatible chip, HID-KBW mode). Also reproduced on **Yokoscan EP8280J** handheld scanner.
- Fleet: 24 production access-control readers, ~6 months uptime

### What happens
When a composite HID device drops off the bus abruptly (our scanners do this when they enter deep-sleep and let go of the USB connection), `hid_host_disable_interface()` ends up calling `usb_host_endpoint_halt()` / `usb_host_endpoint_flush()`, which return `ESP_ERR_INVALID_STATE`. The `HID_RETURN_ON_ERROR()` macro then aborts the cleanup chain — `hid_host_uninstall_device()` is never reached and the USB device context stays dangling.

### Why this is worse than a handle leak
The only way we've found to recover is to **physically unplug and re-plug the scanner's USB cable**. An ESP32 software reboot was not needed — WiFi/heartbeat kept running normally. Nothing we tried in software fixed it:

- `hid_host_uninstall()` + re-install
- `usb_host_uninstall()` + re-install
- Polling for a new attach event
- Full HID driver reinit after 3 / 10 / 30 minutes of silence

Only a fresh physical attach cleared whatever state was stuck. Same hardware works for years on Linux USB hosts (BioStar access-control terminals, Raspberry Pi, x86 kiosk PCs) so it's unlikely to be the scanner firmware.

### Repro
1. Connect an IOTC 0370 / Waveshare-based HID-KBW scanner with Sensing mode.
2. Leave it idle ~2 min until it enters sleep.
3. Wait for the bus drop (or unplug briefly).
4. Logs: `usb_host_endpoint_halt` → `ESP_ERR_INVALID_STATE`, `hid_host_disable_interface` returns early, `hid_host_uninstall_device` never runs.
5. No HID CONNECTED event ever arrives again, even after re-enumeration attempts.

Verified on 1.1.0 and 1.2.0.

### What we've already shipped while waiting for a proper fix
Over the last day, to limit downtime in production, we layered in:

- Passive silent-hang detector (reinit if no scan byte for 10 min)
[...]

PR #471, opened 2026-04-14
hid_host: don't abort disconnect cleanup on endpoint_halt / close errors (IEC-521)

munggo:
Addresses #470.

### Summary

On the disconnect path for composite HID devices, the current implementation aborts the cleanup chain on the first error from `usb_host_endpoint_halt` / `usb_host_endpoint_flush` or per-interface close. Since the device is already gone from the bus, these calls typically return `ESP_ERR_INVALID_STATE` — which is expected and non-fatal — but `HID_RETURN_ON_ERROR` exits the function and `hid_host_uninstall_device()` is never reached.

The practical consequence in our deployment was that the HID device context became stale and could only be cleared by **physically re-plugging the USB cable**. `hid_host_uninstall()` / `usb_host_uninstall()` + re-install, repeated enumeration polling, and HID driver reinits all failed to recover the stuck state.

### Change

Two call sites on the disconnect path now log the failure and continue:

- `hid_host_disable_interface`: `usb_host_endpoint_halt` and `usb_host_endpoint_flush` failures are demoted to `ESP_LOGD` and execution continues to `usb_host_endpoint_clear` and the state transition.
- `hid_host_device_disconnected`: per-interface `hid_host_device_close` failures and the final `hid_host_uninstall_device` failure are demoted to `ESP_LOGW` so the full cleanup chain runs.

This matches **Option A** from #470. The graceful-close path (e.g., a user application calling `hid_host_device_close()` followed by `hid_host_uninstall()` on its own) keeps the existing strict error propagation — the lenient handling only applies when these are reached from `hid_host_device_disconnected()` or from `hid_host_disable_interface()` triggered by a gone device.

### Reproduction (from #470)

Tested with AliExpress IOTC 0370 barcode scanner modules (Waveshare Barcode Scanner Module-compatible chip) and Yokoscan EP8280J handheld scanners in HID-KBW mode, on ESP32-S3-WROOM-1-N16R8 + ESP-IDF v5.2.5. The scanners enter deep sleep when idle and eventually drop off the bus, reliably triggering this path.

Steps:

1. Connect an IOTC 0370 / Waveshare-based HID-KBW scanner.
2. Leave it idle ~2 minutes so it enters internal sleep.
3. Wait for the bus-drop (or unplug briefly to simulate).
4. Without this patch, `usb_host_endpoint_halt` returns `ESP_ERR_INVALID_STATE`, `hid_host_disable_interface` exits via `HID_RETURN_ON_ERROR`, and no further HID CONNECTED event is ever delivered.

### Production experience

A local version of this patch (against v1.1.0) has been running in our fleet of 24 access-control readers for roughly six months with no regressions we can attribute to it. Rebased onto v1.2.0 `master` for this PR.

### Open items

- Happy to shape this toward Option B (opt-in flag) or Option C (split graceful / disconnect paths) if the maintainers prefer.
- A `test_app/` repro that simulates `endpoint_halt` failure mid-cleanup via `hid_mock_device` is not yet included — I'd like to discuss the preferred test strategy (mock hook vs. fault injection in `hid_host_disable_interface`) before adding it, so the test matches whatever the final shape of the patch is.

### Checklist

- [x] Only touches the USB HID Host component
- [x] CHANGELOG entry under `[Unreleased]`
- [x] Fixes linked issue
- [ ] `test_app/` case — will add after we agree on direction

---

> [!NOTE]
> **Medium Risk**
> Touches core USB HID host disconnect/close state handling; while intended to improve robustness, incorrect state forcing or cleanup ordering could cause new edge-case leaks or double-frees in rare error paths.
> 
> **Overview**
> Improves **disconnect cleanup robustness** for composite HID devices by no longer aborting the cleanup chain when `usb_host_endpoint_halt()`/`usb_host_endpoint_flush()` or per-interface close steps fail after a device has already disappeared.
> 
> Adds disconnect-specific best-effort helpers (`hid_host_disable_interface_disconnect()`, `hid_host_device_close_disconnect()`) and refactors close logic into `hid_host_device_close_impl(best_effort)` to *log and continue*, force interface state forward to guarantee list removal, and free `iface->in_xfer` if release fails mid-way; the public graceful-close APIs keep strict error propagation unchanged. Also adds a `[Unreleased]` changelog entry referencing issue #470.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d28a8b95ff5471617a359411fd7cc49afda41194. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


munggo:
Thanks @cursor bot — both reviews are valid. Pushed `49e5374` to rework this along **Option C** from the PR description (split paths) instead of Option A (global best-effort) so the graceful close API keeps its existing contract.

### Addressing review 1 — *endpoint errors silenced on all paths*

`hid_host_disable_interface()` is restored to the original strict `HID_RETURN_ON_ERROR` semantics. A new `hid_host_disable_interface_disconnect()` is introduced and is the only entry point whose `usb_host_endpoint_halt` / `usb_host_endpoint_flush` failures are demoted to `ESP_LOGD`. That variant is called only from `hid_host_device_disconnected()` (via `hid_host_device_close_disconnect()`), so `hid_host_device_stop()` and `hid_host_device_close()` once again surface halt/flush failures to the caller.

### Addressing review 2 — *device freed while failed-close interfaces still reference it*

Two layers of defense:

1. `hid_host_device_close_impl()` — shared body behind both `hid_host_device_close` (strict) and the new static `hid_host_device_close_disconnect` (best-effort). In best-effort mode, a failed `hid_host_disable_interface_disconnect()` logs at `ESP_LOGW` and forces `iface->state = HID_INTERFACE_STATE_READY` so the release-and-free-transfer block below still runs and the iface reliably reaches `_hid_host_remove_interface()`. A failed `hid_host_interface_release_and_free_transfer()` is likewise logged and falls through to the state-transition / list-removal code. That closes the primary path that was leaving interfaces in the tailq.

2. Belt-and-braces guard in `hid_host_device_disconnected()`: immediately before calling `hid_host_uninstall_device()`, the function re-walks `hid_ifaces_tailq` and clears the `parent` pointer of any interface still referencing the `hid_device_t` about to be freed. Any later list traversal that reaches such an orphan now sees `parent == NULL` instead of dereferencing freed memory. This covers pathological cases where close failed so early that even the forced state transition couldn't remove the iface.

### Diff summary

- `hid_host_disable_interface()` — restored to v1.2.0 strict behavior (pre-patch).
- `hid_host_disable_interface_disconnect()` — new, best-effort.
- `hid_host_device_close_impl(handle, best_effort)` — new shared body.
- `hid_host_device_close()` — now a thin wrapper `→ _impl(handle, false)`.
- `hid_host_device_close_disconnect()` — new static wrapper `→ _impl(handle, true)`.
- `hid_host_device_disconnected()` — now calls `_close_disconnect()` per interface and walks the tailq once more before `hid_host_uninstall_device()` to null dangling parents.

Graceful-close strictness is verifiable statically by grep: the only callers of `_disconnect` helpers are `hid_host_device_disconnected()` itself and `hid_host_device_close_disconnect()` (which is itself only called from `hid_host_device_disconnected()`).

CHANGELOG entry under `[Unreleased]` updated to reflect the final shape. The `test_app/` reproduction is still pending — same position as before, happy to add it once the approach looks good to maintainers.

munggo:
@cursor bot — pushed `d452396` addressing all three.

### 1. `ESP_LOGW` in critical section (+ spurious warning)

Both concerns collapsed into the same fix: **remove the orphan-parent-clearing loop entirely.** On reflection it was both wrong and unnecessary:

- *Wrong:* `HID_INTERFACE_STATE_WAIT_USER_DELETION` is a **normal** resting state — an interface with a user callback intentionally sits in `hid_ifaces_tailq` between the first `hid_host_device_close()` (driver-initiated, from the disconnect loop) and the second one (user-initiated, from the `HID_HOST_INTERFACE_EVENT_DISCONNECTED` handler). Clearing `parent` there would break the user's subsequent close, which is exactly the right concern.
- *Unnecessary:* the `best_effort` branch in `hid_host_device_close_impl()` already forces `iface->state` forward on disable/release failure, so the release-and-remove block below always runs and the iface reliably leaves the tailq before `hid_host_uninstall_device()` frees the owning `hid_device_t`. There is no reachable state under which a hid_device is freed while a non-WAIT_USER_DELETION iface still references it.

The critical-section logging concern therefore also goes away (no more `ESP_LOGW` inside `HID_ENTER_CRITICAL()`), and the code is simpler. Left a comment at the spot explaining why the defensive loop isn't needed so it doesn't get re-added by a future reader.

### 2. Transfer buffer leak on release failure (best-effort path)

When `usb_host_interface_release()` returns an error inside `hid_host_interface_release_and_free_transfer()`, that function returns before `usb_host_transfer_free(iface->in_xfer)`. The best-effort caller previously just logged and continued, eventually reaching `_hid_host_remove_interface()` which frees the iface struct and loses the only pointer to `in_xfer` — a genuine leak.

In `hid_host_device_close_impl()`'s best-effort branch, after logging the release failure, I now:

1. Call `usb_host_transfer_free(iface->in_xfer)` directly (if non-null), log on its own failure, and null the pointer. This matches the strict path's invariant: "by the time the iface is removed, `in_xfer` is freed."
2. Force `iface->state = HID_INTERFACE_STATE_IDLE` so a user's subsequent second close doesn't re-enter the release branch on an iface whose transfer is already gone. Without this the iface would remain in `HID_INTERFACE_STATE_READY` and the next close would call `hid_host_interface_release_and_free_transfer()` again on a double-free'd xfer pointer.

CHANGELOG updated to mention the `in_xfer` cleanup.

munggo:
**Correction on deployment scale**

In the original PR description I wrote "a fleet of 24 access-control readers". That number was inaccurate. Actual deployment scale at the time of this PR is:

- **1 unit in production** — a single ESP32-S3 + IOTC 0370 scanner at our Seoul office 3F access gate, running a local version of this patch (against v1.1.0 originally, now 1.2.0) for several weeks.
- **5 units on the bench (dev environment)** — used for firmware / cleanup-chain regression testing.

The silent-hang symptom and the fix motivation stand — we observed the issue reliably on the single production unit over days and reproduced it on the bench units. But I wanted to correct the scale so reviewers aren't misled about battle-test coverage. Extra scrutiny on the Option C split-paths approach (commit `d452396`) is very welcome.

tore-espressif (approved):
@munggo LGTM, thanks!

I'll rebase you PR on latest master where our CI is fixed and if all pass, we can merge

PR #488, opened 2026-05-02
fix(cdc/ftdi): correct divisor selection in 1 MBaud to 2 MBaud range (IEC-534)

spectrumjade:
fix(cdc/ftdi): correct divisor selection in 1 MBaud to 2 MBaud range

## Description
Previously, baud rates between above 1 MBaud were short-circuited to a special case divisor (either 0 or 1), but had the wrong reported realized rate. In the case of `baudrate >= 1000000`, the selected divisor was 1, which programs the chip at 2 MBaud (per AN232B-05), not 1 MBaud as claimed in the comment and set as `baudrate_real`. In addition, rates that the general divisor path could have handled (notably including 1.5 MBaud, which uses even divisor of 2) were being silently coerced to 2 MBaud.

This PR corrects the two special case divisors and allows baud rates below 2 MBaud to be handled by the general divisor path. This allows 1.5 MBaud to work properly.

## Testing
Confirmed that an FT232R, which previously was not successfully able to function at a requested 1.5 MBaud, is now working properly at 1.5 MBaud.

---

## Checklist

### DIFF
--- a/host/class/cdc/usb_host_ftdi_vcp/usb_host_ftdi_vcp.c
+++ b/host/class/cdc/usb_host_ftdi_vcp/usb_host_ftdi_vcp.c
@@ -136,16 +136,16 @@ static void ftdi_event(const cdc_acm_host_dev_event_data_t *event, void *user_ct
 static int calculate_baudrate(uint32_t baudrate, uint16_t *wValue, uint16_t *wIndex)
 {
     int baudrate_real;
-    if (baudrate > 2000000) {
-        // set to 3000000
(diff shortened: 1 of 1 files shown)


tore-espressif:
@spectrumjade thank you for the contribution!



I'll have to check why we did not enable 1.5Mbit. IIRC, the chips that I tested with did not support this baudrate. I'll double check so we can merge this PR, thanks!

tore-espressif (approved):
@spectrumjade thanks again!

PR #489, opened 2026-05-02
fix(uvc): Provide default URB size based on MPS

tore-espressif:
Previously we checked user provided URB size against dwMaxPayloadTransferSize. However, some cameras could return unreasonably large or small values, limiting configuration flexibility.
Now we provide a default URB size of 4x MPS, which is a reasonable starting point.

Changes https://github.com/espressif/esp-usb/pull/450

### DIFF
--- a/host/class/uvc/usb_host_uvc/CHANGELOG.md
+++ b/host/class/uvc/usb_host_uvc/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Fixed regression from version 2.5.0 where unreasonably large URBs could be allocated, leading to out-of-memory scenarios. Now, if urb_size == 0, we allocate URB of size 4x MPS. For other urb_sizes, user's value is used to provide flexibility.
+
 ## [2.5.0] - 2026-04-13
 
 ### Added
--- a/host/class/uvc/usb_host_uvc/include/usb/uvc_host.h
+++ b/host/class/uvc/usb_host_uvc/include/usb/uvc_host.h
@@ -199,7 +199,7 @@ typedef struct {
         size_t frame_size;           /*!< Frame buffer size. Use 0 to take dwMaxVideoFrameSize from negotiation. */
         uint32_t frame_heap_caps;    /*!< Memory capabilities for frame buffers passed to heap_caps_malloc(). */
         int number_of_urbs;          /*!< Number of URBs used by this stream. Triple buffering is recommended. */
-        size_t urb_size;             /*!< Size in bytes of one URB. Larger values trade memory for fewer interrupts. */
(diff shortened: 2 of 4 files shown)


host/class/uvc/usb_host_uvc/uvc_host.c
```diff
@@ -374,6 +374,15 @@ static esp_err_t uvc_transfers_allocate(uvc_stream_t *uvc_stream, unsigned num_o
     if (is_isoc) {
         // Multiply MPS by number of transactions in microframe: This is the minimum size we can request in IN transfer
         max_packet_size *= (USB_EP_DESC_GET_MULT(ep_desc) + 1);
+    }
+
+    if (transfer_size == 0) {
+        // If the caller doesn't specify transfer size, we will allocate 4 MPS buffer
+        // This is a reasonable compromise between performance and memory consumption
+        transfer_size = 4 * max_packet_size;
```
TDA-2030, line 382: I'm a bit curious. What's the most likely actual size of 4*MPS in real-world testing? Is it 2048 bytes?

TDA-2030 (approved):
LGTM

host/class/uvc/usb_host_uvc/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+## Fixed
```
igi540:
```suggestion
### Fixed
```
In Keep a Changelog format this should be ### Fixed; otherwise release tooling or readers may interpret Fixed as a separate top-level release section

host/class/uvc/usb_host_uvc/include/usb/uvc_host.h
```diff
@@ -199,7 +199,7 @@ typedef struct {
         size_t frame_size;           /*!< Frame buffer size. Use 0 to take dwMaxVideoFrameSize from negotiation. */
         uint32_t frame_heap_caps;    /*!< Memory capabilities for frame buffers passed to heap_caps_malloc(). */
         int number_of_urbs;          /*!< Number of URBs used by this stream. Triple buffering is recommended. */
-        size_t urb_size;             /*!< Size in bytes of one URB. Larger values trade memory for fewer interrupts. */
+        size_t urb_size;             /*!< Size in bytes of one URB. Larger values trade memory for fewer interrupts. Set to 0 to use the default size, which is 4x MPS*/
```
igi540:
```suggestion
        size_t urb_size;             /*!< Size in bytes of one URB. Larger values trade memory for fewer interrupts. Set to 0 to use the default size, which is 4x MPS */
```

igi540 (approved):

host/class/uvc/usb_host_uvc/uvc_host.c
tore-espressif: Unfortunately, the interval is pretty big. 

For example, Full Speed Bulk camera will have MPS 64. 64x4 = 256 bytes

A High Speed Isochronous camera with 2 extra transactions in a microframe will have MPS 3x1024. 3x1024x4 = 12288 bytes.

So this is really just a starting point that should always work for the user. If the users needs to fine-tune memory consumption and performance, he will have to modify the value...

tore-espressif:
@igi540 @TDA-2030 Thank you for your reviews, will fix ASAP!

PR #501, opened 2026-06-03
Remove idf5.4 usb managed support

tore-espressif:
Overriding esp-idf native `usb` component with managed component from esp-usb was supported for all service period releases. At the time of first release, this was 5.4 and 5.5

IDF 5.4 entered maintenance period in January 2026. It is time to drop support for managed usb in5.4

The minimal required version is now 5.5.3 which already support remote wake-up HAL API

### DIFF
--- a/host/usb/src/hcd_dwc.c
+++ b/host/usb/src/hcd_dwc.c
@@ -863,7 +863,6 @@ static hcd_port_event_t _intr_hdlr_hprt(port_t *port, usb_dwc_hal_port_event_t h
         port->flags.conn_dev_ena = 0;
         break;
     }
-#ifdef REMOTE_WAKE_HAL_SUPPORTED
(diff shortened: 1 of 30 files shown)


host/usb/src/hcd_dwc.c
peter-marcisovsky: We can delete this note

peter-marcisovsky (approved):
All the target tests still use `if: "idf_version >=5.4"`, for [example](https://github.com/espressif/esp-usb/blob/master/host/usb/test/target_test/common/idf_component.yml#L7). Should we keep the supported IDF for the tests as it is?

host/usb/src/hcd_dwc.c
tore-espressif: Thanks and done.

I also removed this note from all our usb header file

``` c
/*
Warning: The USB Host Library API is still a beta version and may be subject to change
*/
```

PR #375, opened 2026-01-13
fix(host_test): Wrap main for argc and argv

tore-espressif:
All host tests are updated so
1. `main()` is wrapped, so the esp-idf/linux implementation is not called
2. in `__wrap_main()` we do the same as in the original implementation, but we also pass argc and argv arguments to `main_task` and then to Catch 2

This has great advantages in debugging, because we can now pass CLI arguments to the Catch2 application. Eg.

``` sh
$ ./build/host_test_usb_uvc.elf --help

Catch2 v3.7.0
usage:
  host_test_usb_uvc.elf [<test name|pattern|tags> ... ] options

where options are:
  -?, -h, --help                            display usage information
  -s, --success                             include successful tests in
                                            output
  -b, --break                               break into debugger on failure
...
```
or
``` sh
$ ./build/host_test_usb_uvc.elf --list-tags
All available tags:
   1  [anker]
   1  [bulk]
   1  [c200]
   1  [c270]
   1  [c980]
   1  [canyon]
...
```
or
``` sh
$ ./build/host_test_usb_uvc.elf [logitech]
Filters: [logitech]
Randomness seeded to: 3864178348
===============================================================================
All tests passed (2928 assertions in 3 test cases)
```

PR #503, opened 2026-06-04
change: Do not wrap main in linux tests

tore-espressif:
Latest IDF provides weak main, so we do not need to wrap

Partly reverts https://github.com/espressif/esp-usb/pull/375 because esp-idf now supports main overriding without linker wrapping.

Also fixes build on MacOS

### DIFF
--- a/host/class/cdc/usb_host_cdc_acm/host_test/device_interaction/main/test_main.cpp
+++ b/host/class/cdc/usb_host_cdc_acm/host_test/device_interaction/main/test_main.cpp
@@ -5,6 +5,7 @@
  */
 
 #include <stdio.h>
+#include <unistd.h>
 #include <catch2/catch_session.hpp>
 #include "freertos/FreeRTOS.h"
 #include "freertos/task.h"
@@ -25,7 +26,11 @@ static void main_task(void *args)
     vTaskDelete(NULL);
 }
 
-extern "C" int __wrap_main(int argc, const char **argv)
+extern "C" void app_main(void)
+{
+}
+
+int main(int argc, const char **argv)
 {
     // Following section is copied from components\freertos\FreeRTOS-Kernel\portable\linux\port_idf.c
     // It starts the FreeRTOS scheduler and creates the main task to run Catch2 tests.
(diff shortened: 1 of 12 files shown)


tore-espressif:
@peter-marcisovsky 1 change, PTAL
1. in esp-idf, `main()` for linux target is now defined as weak. So we do not need the `--wrap` argument for linker. Moreover, the `--wrap` argument does not work on MacOS. Other than that, the host tests run the same on linux and MacOS

host/class/cdc/usb_host_cdc_acm/host_test/device_interaction/main/test_main.cpp
```diff
@@ -25,7 +26,11 @@ static void main_task(void *args)
     vTaskDelete(NULL);
 }
 
-extern "C" int __wrap_main(int argc, const char **argv)
+extern "C" void app_main(void)
```
peter-marcisovsky, line 29: Is the `app_main` just for a purpose of a "completeness" of having an app_main declared in a project?
For me, on linux, the test host test builds even without having the `app_main` declared in the project.

peter-marcisovsky (approved):
Cool update. LGTM.

host/class/cdc/usb_host_cdc_acm/host_test/device_interaction/main/test_main.cpp
tore-espressif: I'll double check.

tore-espressif: Indeed, this symbol is required for arm64 builds...

BTW, can you try it on linux again from clean state? (When I removed `app_main` on x86 linux _re-build_ succeeded, but `idf.py fullclean build` failed...)

tore-espressif: On x86 architecture and WSL linux, it works even without `app_main` definion...

So I'll merge it as-is

PR #515, opened 2026-06-24
feat(usb_host): add existing device enumeration

TDA-2030:
## Problem Solved

  - When registering multiple classes using a composite device, the registration of different classes may occur at different times. If the enumeration of the USB device takes place prior to class registration, the subsequently registered classes will never receive the device connection event until re-enumeration is performed.

## Changes

  - Add `enumerate_existing_devices` to USB Host client config so late-registered clients can opt in to receiving `NEW_DEV`
  events for already visible devices.
  - Track client-visible device addresses in the USB Host layer and keep real `NEW_DEV` delivery ordered with historical event
  replay.
  - Skip client-facing remove/gone events for default address 0 devices that disappear during early enumeration.

## Testing

  - Add USB Host timing target tests for early/late clients, reconnects, opened/unopened clients, opt-in replay, and stale-
  device cleanup.

---

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes core client event ordering and default behavior for late registration (more NEW_DEV delivery), with new locking paths around pending replay; mistakes could cause duplicate/missed discovery or races under queue backpressure.
> 
> **Overview**
> **Late-registered USB host clients now get `NEW_DEV` for devices the stack already enumerated**, fixing composite setups where class drivers register after connect. The host layer tracks **client-visible** device addresses and, on each new device, marks them visible and queues delivery per client instead of a single broadcast that could be lost when an event queue is full.
> 
> Delivery uses **pending bitmasks** per client plus a **`has_pending_new_dev`** flag; `usb_host_client_handle_events` retries after the queue drains. **Registration** seeds pending from the global visible map and flushes immediately. **Removal** clears visible/pending state (including on `usb_host_device_free_all`), and **address 0** no longer generates client remove/gone traffic during early enumeration.
> 
> Opened-device tracking is refactored to shared **`dev_addr_map_*`** helpers. **Target timing tests** cover early/late/mixed clients, reconnects, `DEV_GONE`/`DEV_REMOVED` routing, free-all, and in-flight disconnect.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a4d7f1a557947a4db4d2c568bcbf0e0f026a1ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/host/usb/src/usb_host.c
+++ b/host/usb/src/usb_host.c
@@ -62,8 +62,12 @@ DEFINE_CRIT_SECTION_LOCK_STATIC(host_lock);
 #define PROCESS_REQUEST_PENDING_FLAG_HUB        (1 << 1)
 #define PROCESS_REQUEST_PENDING_FLAG_ENUM       (1 << 2)
 
+#define DEV_ADDR_MAP_LEN                        4
+
 #define SHORT_DESC_REQ_LEN                      8
 #define CTRL_TRANSFER_MAX_DATA_LEN              CONFIG_USB_HOST_CONTROL_TRANSFER_MAX_SIZE
+
+typedef uint32_t dev_addr_map_t[DEV_ADDR_MAP_LEN];
 
 typedef struct ep_wrapper_s ep_wrapper_t;
 typedef struct interface_s interface_t;
@@ -114,18 +118,20 @@ struct client_s {
             struct {
                 uint32_t handling_events: 1;
                 uint32_t taking_mux: 1;
-                uint32_t reserved6: 6;
+                uint32_t has_pending_new_dev: 1;
+                uint32_t reserved5: 5;
                 uint32_t num_intf_claimed: 8;
                 uint32_t reserved16: 16;
             };
             uint32_t val;
         } flags;
         uint32_t num_done_ctrl_xfer;
-        uint32_t opened_dev_addr_map[4];
+        dev_addr_map_t opened_dev_addr_map;
     } dynamic;
     // Mux protected members must be protected by host library the mux_lock when accessed
     struct {
         TAILQ_HEAD(tailhead_interfaces, interface_s) interface_tailq;
+        dev_addr_map_t pending_new_dev_addr_map;
     } mux_protected;
     // Constant members do no change after registration thus do not require a critical section
     struct {
@@ -156,6 +162,8 @@ typedef struct {
     // Mux protected members must be protected by host library the mux_lock when accessed
     struct {
         TAILQ_HEAD(tailhead_clients, client_s) client_tailq;  // List of all clients registered
+        // Devices whose NEW_DEV event has already been processed by the host lib
+        dev_addr_map_t client_visible_dev_addr_map;
     } mux_protected;
     // Constant members do no change after installation thus do not require a critical section
     struct {
@@ -174,28 +182,46 @@ const char *USB_HOST_TAG = "USB HOST";
 
 // ----------------------------------------------------- Helpers -------------------------------------------------------
 
-static inline void _record_client_opened_device(client_t *client_obj, uint8_t dev_addr)
+static inline void dev_addr_map_set(dev_addr_map_t dev_addr_map, uint8_t dev_addr)
 {
     assert(dev_addr != 0 && dev_addr <= 127);
-    client_obj->dynamic.opened_dev_addr_map[dev_addr / 32] |= (uint32_t)(1 << (dev_addr % 32));
+    dev_addr_map[dev_addr / 32] |= (1U << (dev_addr % 32));
 }
 
-static inline void _clear_client_opened_device(client_t *client_obj, uint8_t dev_addr)
+static inline void dev_addr_map_clear(dev_addr_map_t dev_addr_map, uint8_t dev_addr)
 {
     assert(dev_addr != 0 && dev_addr <= 127);
-    client_obj->dynamic.opened_dev_addr_map[dev_addr / 32] &= ~(uint32_t)(1 << (dev_addr % 32));
+    dev_addr_map[dev_addr / 32] &= ~(1U << (dev_addr % 32));
 }
 
-static inline bool _check_client_opened_device(client_t *client_obj, uint8_t dev_addr)
+static inline bool dev_addr_map_check(dev_addr_map_t dev_addr_map, uint8_t dev_addr)
 {
-    bool ret;
     assert(dev_addr <= 127);
-    if (dev_addr != 0) {
-        ret = client_obj->dynamic.opened_dev_addr_map[dev_addr / 32] & (uint32_t)(1 << (dev_addr % 32));
-    } else {
-        ret = false;
+    return (dev_addr != 0) ? ((dev_addr_map[dev_addr / 32] & (1U << (dev_addr % 32))) != 0) : false;
+}
+
+static inline void dev_addr_map_copy(dev_addr_map_t dst, const dev_addr_map_t src)
+{
+    for (size_t i = 0; i < DEV_ADDR_MAP_LEN; i++) {
+        dst[i] = src[i];
     }
-    return ret;
+}
+
+static inline void dev_addr_map_clear_all(dev_addr_map_t dev_addr_map)
+{
+    for (size_t i = 0; i < DEV_ADDR_MAP_LEN; i++) {
+        dev_addr_map[i] = 0;
+    }
+}
+
+static inline bool dev_addr_map_any(const dev_addr_map_t dev_addr_map)
+{
+    for (size_t i = 0; i < DEV_ADDR_MAP_LEN; i++) {
+        if (dev_addr_map[i] != 0) {
+            return true;
+        }
+    }
+    return false;
 }
 
 static bool _unblock_client(client_t *client_obj, bool in_isr)
@@ -269,6 +295,21 @@ static void stop_auto_suspend_timer(void)
     xTimerStop(p_host_lib_obj->constant.auto_suspend_timer, portMAX_DELAY);
 }
 
+static bool send_event_msg_to_client(client_t *client_obj, const usb_host_client_event_msg_t *event_msg, const char *error_msg)
+{
+    // The caller must hold mux_lock when calling this helper.
+    if (xQueueSend(client_obj->constant.event_msg_queue, event_msg, 0) == pdTRUE) {
+        HOST_ENTER_CRITICAL();
+        _unblock_client(client_obj, false);
+        HOST_EXIT_CRITICAL();
+        return true;
+    }
+    if (error_msg) {
+        ESP_LOGE(USB_HOST_TAG, "%s", error_msg);
+    }
+    return false;
+}
+
 static void send_event_msg_to_clients(const usb_host_client_event_msg_t *event_msg, bool send_to_all, uint8_t opened_dev_addr)
 {
     // Lock client list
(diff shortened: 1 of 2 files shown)


TDA-2030:
@tore-espressif Should we set `enumerate_existing_devices` to true by default?

host/usb/src/usb_host.c
```diff
@@ -198,6 +200,18 @@ static inline bool _check_client_opened_device(client_t *client_obj, uint8_t dev
     return ret;
 }
 
+static inline void _record_client_visible_device(uint8_t dev_addr)
```
peter-marcisovsky:
```suggestion
static inline void record_client_visible_device(uint8_t dev_addr)
```
We use `_` prefix for functions called from critical sections. I would suggest to remove the prefix, as the function is not required to be called from critical section. 

For the functions required to be called with mutex locked, we don't have any specific nomenclature. Usually just put a comment like [here](https://github.com/espressif/esp-usb/blob/master/host/usb/src/usbh.c#L219-L221) 

Same for the below function.

tore-espressif:
@TDA-2030 doesn't this function already solve your problem? 
https://github.com/espressif/esp-usb/blob/de5d4a19cee0dcb9c33866af2eb2014a8e30d5a0/host/usb/include/usb/usb_host.h#L459

TDA-2030:
> @TDA-2030 doesn't this function already solve your problem?
> 
> https://github.com/espressif/esp-usb/blob/de5d4a19cee0dcb9c33866af2eb2014a8e30d5a0/host/usb/include/usb/usb_host.h#L459

This API requires constant polling to detect new devices. In my opinion, event-based notification for device attach/detach fits USB hotplug semantics much better. The `USB_HOST_CLIENT_EVENT_NEW_DEV` event is already defined, and this fix completes the missing event dispatch.

I’d appreciate your feedback if you have any concerns or different opinions.

host/usb/src/usb_host.c
```diff
@@ -285,19 +311,58 @@ static void send_event_msg_to_clients(const usb_host_client_event_msg_t *event_m
...
+{
+    const usb_host_client_event_msg_t event_msg = {
+        .event = USB_HOST_CLIENT_EVENT_NEW_DEV,
+        .new_dev.address = dev_addr,
+    };
+
+    xSemaphoreTake(p_host_lib_obj->constant.mux_lock, portMAX_DELAY);
+    client_t *client_obj;
+    TAILQ_FOREACH(client_obj, &p_host_lib_obj->mux_protected.client_tailq, dynamic.tailq_entry) {
+        send_event_msg_to_client(client_obj, &event_msg, "Client event message queue full");
```
peter-marcisovsky: If a client queue is full, the event is dropped but the device is still marked visible. You could reconsider handling of the error message here.

host/usb/src/usb_host.c
```diff
@@ -285,19 +311,58 @@ static void send_event_msg_to_clients(const usb_host_client_event_msg_t *event_m
...
+    // Mark the device visible while holding mux_lock so later client registration cannot miss this device.
+    _record_client_visible_device(dev_addr);
+    xSemaphoreGive(p_host_lib_obj->constant.mux_lock);
+}
+
+static void send_existing_new_dev_events_to_client(client_t *client_obj)
+{
+    // The caller must hold mux_lock to keep historical NEW_DEV delivery ordered with real enumeration.
+    for (uint8_t dev_addr = 1; dev_addr <= 127; dev_addr++) {
+        if ((p_host_lib_obj->mux_protected.client_visible_dev_addr_map[dev_addr / 32] & (1U << (dev_addr % 32))) == 0) {
```
peter-marcisovsky: This logical expression is bit hard to decode, could you consider adding some comment what is happening here? Or create an inline function from which name it would be more clear, just like `_clear_client_visible_device` for example.

host/usb/src/usb_host.c
```diff
@@ -285,19 +311,58 @@ static void send_event_msg_to_clients(const usb_host_client_event_msg_t *event_m
...
+    }
+    // Mark the device visible while holding mux_lock so later client registration cannot miss this device.
+    _record_client_visible_device(dev_addr);
+    xSemaphoreGive(p_host_lib_obj->constant.mux_lock);
+}
+
+static void send_existing_new_dev_events_to_client(client_t *client_obj)
+{
+    // The caller must hold mux_lock to keep historical NEW_DEV delivery ordered with real enumeration.
+    for (uint8_t dev_addr = 1; dev_addr <= 127; dev_addr++) {
```
peter-marcisovsky, line 338: There could be up to 127 events `USB_HOST_CLIENT_EVENT_NEW_DEV` enqueued to the client. Even thought, there is a `max_num_event_msg` when registering client, this function does not validate the `max_num_event_msg` parameter. 

The concern here is to have more devices connected than `max_num_event_msg`. So the client could miss a device enumeration.

host/usb/src/usb_host.c
```diff
@@ -1189,13 +1261,17 @@ esp_err_t usb_host_client_register(const usb_host_client_config_t *client_config
...
     client_obj->constant.notify_dev_removed = client_config->flags.notify_dev_removed;
+    client_obj->constant.enumerate_existing_devices = client_config->flags.enumerate_existing_devices;
 
     // Add client to the host library's list of clients
     xSemaphoreTake(p_host_lib_obj->constant.mux_lock, portMAX_DELAY);
     HOST_ENTER_CRITICAL();
     p_host_lib_obj->dynamic.flags.num_clients++;
     HOST_EXIT_CRITICAL();
     TAILQ_INSERT_TAIL(&p_host_lib_obj->mux_protected.client_tailq, client_obj, dynamic.tailq_entry);
+    if (client_obj->constant.enumerate_existing_devices) {
```
peter-marcisovsky: The `enumerate_existing_devices` relies on the connected devcies bitmap, not considering `usbh_devs_addr_list_fill()`. 

A client registering while a device disconnect is in flight (serialized by mux_lock) may get a `NEW_DEV` for a device that is already going away. That is similar to receiving a live NEW_DEV just before unplug.

This not blocker, but maybe worth testing.

host/usb/test/target_test/usb_host/main/test_usb_host_client_timing.c
peter-marcisovsky, line 1: Nice tests, it would be worth adding a test where the `max_num_event_msg` gets tested, thus you would queue more messages than the client could handle.

host/usb/src/usb_host.c
TDA-2030: Fixed.

The direct bitmap expressions were replaced with small dev_addr_map_* helpers. The same helpers are now used for the opened-device map, the globally visible-device map, and each client's pending existing-device map, so the intent should be clearer at the call sites.

host/usb/src/usb_host.c
TDA-2030: Thanks, fixed.

I removed the underscore-style helper naming from this path.

host/usb/src/usb_host.c
TDA-2030: Good point. I changed the handling so the global visible-device bitmap only represents devices that the host library has made visible to the client layer, not whether every individual client has already consumed the event.

For opt-in existing-device enumeration, each client now gets its own pending bitmap snapshot. If the client's event queue is full, the pending bit is kept and delivery is retried from usb_host_client_handle_events(). So the historical NEW_DEV event is no longer lost because of a temporarily full queue.

host/usb/src/usb_host.c
TDA-2030: This is handled differently now.

The implementation no longer requires max_num_event_msg to be large enough to hold all currently visible devices. During registration, the client snapshots the visible-device bitmap into its own pending bitmap. Events are then queued opportunistically, and any device that does not fit into the queue remains pending for a later retry from usb_host_client_handle_events().

Because of that, a client should not miss existing-device enumeration just because there are more devices than max_num_event_msg.

host/usb/src/usb_host.c
TDA-2030: Agreed.

The current implementation keeps this serialized with the host library mutex. Device removal clears the global visible-device bitmap and also clears the same device from every client's pending existing-device bitmap, so a later retry will not replay a stale historical NEW_DEV for a removed device.

An already queued NEW_DEV is intentionally not withdrawn. In that case the client may observe NEW_DEV and then fail to open the device or receive the corresponding removal path, which matches the normal live disconnect behavior.

I also added a timing test that covers opt-in registration while removal is in flight.

host/usb/test/target_test/usb_host/main/test_usb_host_client_timing.c
TDA-2030: I considered adding a target test that overflows max_num_event_msg with multiple existing devices, but making that deterministic would require a hub or multiple attached devices.

Instead, the implementation avoids the need for such a large queue by keeping a per-client pending bitmap and retrying delivery from usb_host_client_handle_events(). So I haven't added this kind of test yet

host/usb/src/usb_host.c
```diff
@@ -279,25 +319,79 @@ static void send_event_msg_to_clients(const usb_host_client_event_msg_t *event_m
...
+    TAILQ_FOREACH(client_obj, &p_host_lib_obj->mux_protected.client_tailq, dynamic.tailq_entry) {
+        send_event_msg_to_client(client_obj, &event_msg, "Client event message queue full");
+    }
+    // Mark the device visible while holding mux_lock so later client registration cannot miss this device.
+    dev_addr_map_set(p_host_lib_obj->mux_protected.client_visible_dev_addr_map, dev_addr);
+    xSemaphoreGive(p_host_lib_obj->constant.mux_lock);
+}
+
+static void send_pending_existing_new_dev_events_to_client(client_t *client_obj)
+{
```
peter-marcisovsky, line 336: Would it make sense to add a `dev_addr_map_any` in the beginning of this function, to see if there are any pending device addresses in `pending_existing_dev_addr_map`. If there are none, return early. If there are, continue. 

Instead of traversing all 127 addresses and checking each one?

host/usb/src/usb_host.c
```diff
@@ -174,28 +181,46 @@ const char *USB_HOST_TAG = "USB HOST";
...
-static inline void _clear_client_opened_device(client_t *client_obj, uint8_t dev_addr)
+static inline void dev_addr_map_clear(dev_addr_map_t dev_addr_map, uint8_t dev_addr)
 {
     assert(dev_addr != 0 && dev_addr <= 127);
-    client_obj->dynamic.opened_dev_addr_map[dev_addr / 32] &= ~(uint32_t)(1 << (dev_addr % 32));
+    dev_addr_map[dev_addr / 32] &= ~(1U << (dev_addr % 32));
 }
 
-static inline bool _check_client_opened_device(client_t *client_obj, uint8_t dev_addr)
+static inline bool dev_addr_map_test(dev_addr_map_t dev_addr_map, uint8_t dev_addr)
```
peter-marcisovsky: `check` instead of `test` might be better fit here, not to emphasize testing. Up to you.
```suggestion
static inline bool dev_addr_map_check(dev_addr_map_t dev_addr_map, uint8_t dev_addr)
```

host/usb/src/usb_host.c
```diff
@@ -1319,6 +1421,10 @@ esp_err_t usb_host_client_handle_events(usb_host_client_handle_t client_hdl, Tic
             client_obj->constant.event_callback(&event_msg, client_obj->constant.callback_arg);
         }
 
+        xSemaphoreTake(p_host_lib_obj->constant.mux_lock, portMAX_DELAY);
+        send_pending_existing_new_dev_events_to_client(client_obj);
+        xSemaphoreGive(p_host_lib_obj->constant.mux_lock);
```
peter-marcisovsky: The mutex is being acquired during each run of the `usb_host_client_handle_events` which, from point of thread safety is correct, but could affect latency instead. 

There are 2 points that sould be addressed:
- If a client is not registered with `flags.enumerate_existing_devices` the `pending_existing_dev_addr_map` is empty, yet, still the mutex here is taken, `send_pending_existing_new_dev_events_to_client` is called and all the 127 addresses are being traversed with each `usb_host_client_handle_events` run

- if a client is registered with the `flags.enumerate_existing_devices`, but there might be not new device events pending, still the mutex and the function itself is called..


I would suggest to acquire mutex and call the send_pending_existing_new_dev_events_to_client only:
- if the client was registered with `enumerate_existing_devices` flag
- if there are actually any events pending (you can create a flag for that, something like `has_pending_events` which can be read in a critical section above). Just like the usb_host_lib_handle_events function does it.

host/usb/test/target_test/usb_host/main/test_usb_host_client_timing.c
peter-marcisovsky: True, the current CI runner setup is limitation.

FYI, here I have revisited an older linux host test infrastructure idea  
- #520 

Using which it would be possible to test this kind of edge cases in the USB Host layer. I have also added a similar [test](https://github.com/espressif/esp-usb/pull/520/changes#diff-fef09dd4984b506c0fb502436b705b54aef034c4fdec184018794846c2d11d9dR215) which tests the `max_num_event_msg` queue. We can add some more tests for you feature once it's merged.

host/usb/src/usb_host.c
```diff
@@ -279,25 +319,79 @@ static void send_event_msg_to_clients(const usb_host_client_event_msg_t *event_m
...
+        .new_dev.address = dev_addr,
+    };
+
+    xSemaphoreTake(p_host_lib_obj->constant.mux_lock, portMAX_DELAY);
+    client_t *client_obj;
+    TAILQ_FOREACH(client_obj, &p_host_lib_obj->mux_protected.client_tailq, dynamic.tailq_entry) {
+        send_event_msg_to_client(client_obj, &event_msg, "Client event message queue full");
+    }
+    // Mark the device visible while holding mux_lock so later client registration cannot miss this device.
+    dev_addr_map_set(p_host_lib_obj->mux_protected.client_visible_dev_addr_map, dev_addr);
```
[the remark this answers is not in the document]
peter-marcisovsky: The bot's response actually addresses my comment [here](https://github.com/espressif/esp-usb/pull/515#discussion_r3497922754). 

Shouldn't the `send_new_dev_event_msg_to_clients` put a device into a pending list if the client message is full?

host/usb/src/usb_host.c
TDA-2030: Thanks, fixed.

The implementation has changed since the original version: `enumerate_existing_devices` was removed, and NEW_DEV discovery is now reliable by default for every client. Live NEW_DEV and existing-device replay both use the same per-client pending_new_dev_addr_map.

I also added a lightweight `has_pending_new_dev` flag. usb_host_client_handle_events() now reads this flag in a critical section first, and only takes the host library mutex when the client actually has pending NEW_DEV delivery. This avoids acquiring mux_lock on the common no-pending path.

The pending bitmap itself remains protected by mux_lock. If the quick flag is stale after device removal, the next pending drain may acquire the mutex once, observe an empty bitmap, and clear the flag. This keeps the code simple while avoiding repeated mutex acquisition.

host/usb/src/usb_host.c
TDA-2030: Done

host/usb/src/usb_host.c
TDA-2030: I have prevented repeated execution of this function via the `has_pending_new_dev` flag. Meanwhile, I added a judgment on `dev_addr_map_any` at the end of the function to control `has_pending_new_dev`, which avoids repeatedly sending NEW_DEV events.

host/usb/src/usb_host.c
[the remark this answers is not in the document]
TDA-2030: I removed `enumerate_existing_devices`, so both `USBH_EVENT_NEW_DEV` events and events replayed during client registration go through `send_pending_new_dev_events_to_client`. This issue should therefore be resolved.

host/usb/test/target_test/usb_host/main/test_usb_host_client_timing.c
TDA-2030: Yes, adding more tests to linux_test is a more reasonable approach

PR #547, opened 2026-08-21
feat(host): Support suspend/resume in dual host configuration

tore-espressif:
Suspend/Resume both USB host ports in dual host configuration

---

> [!NOTE]
> **Medium Risk**
> Changes core USB host power-management and light-sleep paths for dual-controller setups; incorrect per-device PM handling could affect enumeration or suspend notifications, though new unit tests cover the main scenarios.
> 
> **Overview**
> Fixes **ESP32-P4 dual-host** power management so global suspend, resume, auto-suspend, and pre–light-sleep suspend apply to **every root port with a device**, not only the first controller.
> 
> **Hub driver:** Root ports share one PM state—`hub_root_mark_suspend` / `hub_root_mark_resume`, light-sleep suspend, and `hub_root_is_suspended` iterate all installed ports. Suspend/resume client notifications are issued **once** after both ports are processed. A device attach on an idle port resumes a suspended sibling; light-sleep auto-PM is tracked **per port**.
> 
> **USBH:** `usbh_devs_set_pm_actions_all` only applies suspend/resume **events** to devices already in the matching state, so a new enumeration on one bus is not broken when the other bus resumes.
> 
> Docs, public API comments, changelog, and **Catch2 hub root PM unit tests** (dual-port fake HCD) document and lock in the behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a11abaa723b284ff5e3674801ebbb77483a8aedb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/host/usb/include/usb/usb_host.h
+++ b/host/usb/include/usb/usb_host.h
@@ -96,7 +96,8 @@ typedef struct {
 typedef struct {
     int num_devices;            /**< Current number of connected (and enumerated) devices */
     int num_clients;            /**< Current number of registered clients */
-    bool root_port_suspended;   /**< Current status of the root port (suspended/resumed) */
+    bool root_port_suspended;   /**< Current status of the root ports (suspended/resumed). In dual host configuration,
+                                     this is true only if all the root ports with a connected device are suspended */
 } usb_host_lib_info_t;
 
 // ---------------------- Callbacks ------------------------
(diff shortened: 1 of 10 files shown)


host/usb/include/usb/usb_host.h
```diff
@@ -267,35 +268,41 @@ esp_err_t usb_host_lib_info(usb_host_lib_info_t *info_ret);
...
 /**
- * @brief Suspend the root port
+ * @brief Suspend the root ports
  *
  * - The function checks, if a device is connected and if a transfer is submitted
- * - Then halts and flushes all endpoints of all the connected devices and suspends the root port
+ * - Then halts and flushes all endpoints of all the connected devices and suspends the root ports
  * - Finally, it notifies all the clients which opened the device, that the device is now suspended
  *
  * @note Remote wakeup from device is not implemented yet
```
peter-marcisovsky: This note should be removed

host/usb/src/hub.c
```diff
@@ -1345,20 +1416,26 @@ esp_err_t hub_process(void)
...
 #if HCD_NUM_PORTS > 1
         if (action_flags & HUB_DRIVER_ACTION_ROOT1_EVENT) {
             root_port_handle_events(&p_hub_driver_obj->root_hub_ports[1]);
         }
         if (action_flags & HUB_DRIVER_ACTION_ROOT1_REQ) {
-            root_port_req(&p_hub_driver_obj->root_hub_ports[1]);
+            dev_ctrl |= root_port_req(&p_hub_driver_obj->root_hub_ports[1]);
         }
 #endif // HCD_NUM_PORTS > 1
+        if (dev_ctrl != 0) {
```
peter-marcisovsky, line 1447: Here, any single port success is enough to trigger the global client notification.
In case one of the port finishes the suspend/resume sequence with an error (disconnection during the suspending, or command failed..) and the other port succeeds, the client event is delivered as if both ports were suspended/resumed.


There should be tracking of the port suspend/resume action competition for each port, not just ORed `dev_ctrl` flag.

host/usb/src/hub.c
peter-marcisovsky: This flag is global, but disconnect on one port clears it for the whole hub. `hub_root_mark_exit_light_sleep` then returns early (light_sleep_auto_pm == 0) and skips deferred suspend delivery for the other port.

host/usb/src/hub.c
```diff
@@ -464,9 +509,22 @@ static void root_port_handle_events(root_hub_port_t *root_hub_port)
...
+                sibling_suspended = true;
+                break;
+            }
+        }
         HUB_DRIVER_EXIT_CRITICAL();
+
+        // All root ports share a single PM state. A device attaching to one bus makes the whole Host active,
+        // so any bus that is still suspended must be resumed together with this connection.
+        if (sibling_suspended) {
+            hub_root_mark_resume();
```
peter-marcisovsky: Is this allowed to fail?

host/usb/src/hub.c
```diff
@@ -921,146 +964,165 @@ esp_err_t hub_root_stop(void)
...
+        default:
+            // Port disconnected, in recovery, or otherwise not resumable
+            break;
+        }
+    }
+
     esp_err_t ret;
-    switch (port_state) {
-    case HCD_PORT_STATE_SUSPENDED:
+    if (any_resumable) {
```
peter-marcisovsky, line 1080: In case one port is in resuming state (or recovery) and the other port is suspended. The return always return ESP_OK. and `usb_host_lib_root_port_resume()` proceeds to both port resume. 

Which does not cause hard fail, due to [HCD check](https://github.com/espressif/esp-usb/blob/master/host/usb/src/hcd_dwc.c#L1408), just return error. Which makes it a redundant command to HCD.

You can add HCD state check to hub_root_mark_resume() to issue `PORT_REQ_` only if the HCD is in correct state.

host/usb/test/host_test/usbh_layer_test/main/hub_root_pm_unit_test.cpp
peter-marcisovsky, line 1: Low effort test update: Add remote wakeup event injection and expect both port to be resumed by the remote wake. Seems like just: 
- calling `port_event(1, HCD_PORT_EVENT_REMOTE_WAKEUP);` 
- expecting `REQUIRE_FALSE(hub_root_is_suspended());`

Also worth testing error paths in the implementation. Like suspend both ports, but one command fails and see how the code behaves, as it's bit tricky to simulate this scenario in the real world.

host/usb/test/host_test/usbh_layer_test/main/hub_root_pm_unit_test.cpp
```diff
@@ -0,0 +1,450 @@
...
+        THEN("The Host can neither be suspended nor resumed") {
+            REQUIRE(ESP_ERR_NOT_ALLOWED == hub_root_can_suspend());
+            REQUIRE(ESP_ERR_NOT_ALLOWED == hub_root_mark_suspend());
+            REQUIRE(ESP_ERR_NOT_ALLOWED == hub_root_can_resume());
+            REQUIRE(ESP_ERR_NOT_ALLOWED == hub_root_mark_resume());
+            REQUIRE_FALSE(hub_root_is_suspended());
+            REQUIRE(s_hcd.commands.empty());
+        }
+
+        uninstall_hub(0);
```
peter-marcisovsky:
```suggestion
        uninstall_hub(BIT0);
```

host/usb/src/hub.c
tore-espressif: I added an error log so the error condition is not silent.

However, we are not aborting on the error. The new device and its root port should continue working normally

host/usb/src/hub.c
tore-espressif: `light_sleep_auto_pm` flag is now per root port. (previously 1 per HUB driver)

tore-espressif (commented):
@peter-marcisovsky Thank you very much for the review!

I fixed some of the issues. However, the comments about 1 port failing to suspend/resume are still unresolved

host/usb/test/host_test/usbh_layer_test/main/hub_root_pm_unit_test.cpp
tore-espressif: - [x] Done: Remote wake-up injection
- [ ] TODO: error path tests

PR #583, opened 2026-09-15
fix(host/msc): fix READ CAPACITY(10) off-by-one 

omnitex:
## Description
READ CAPACITY(10) returns the LBA of the **last** accessible logical block, not a block count (SCSI spec). This raw value was stored and used as-is everywhere, under-reporting device capacity by one block (e.g. 60125183 vs actual 60125184 sectors on a 32GB drive).

**Fix:** Convert `last_lba + 1` once in `scsi_cmd_read_capacity()`, so all consumers (`msc_host.c` disk info, `msc_bdl.c` disk_size, `diskio_usb.c` GET_SECTOR_COUNT) get the correct value from a single source of truth. Also explicitly reject the `0xFFFFFFFF` sentinel (capacity requires READ CAPACITY(16), unsupported) instead of silently treating it as a valid LBA.

## Related

See e.g. https://linux.die.net/man/8/sg_readcap

> The SCSI READ CAPACITY command (both 10 and 16 byte cdbs) actually yield the block address of the last block and the block size. The number of blocks is thus one plus the block address of the last block (as blocks are counted origin zero (i.e. starting at block zero)). This is the source of many "off by one" errors. 

## Testing

Verified the off-by-one this way:

For a 32GB USB drive, the `esp-usb/host/class/msc/usb_host_msc/test_app`

`(13)	"device_info" [usb_msc]` test reports

```
Device info:
	 Capacity: 29357 MB
	 Sector size: 512
	 Sector count: 60125183 <<--
	 PID: 0x5595
	 VID: 0x 781
	 iProduct: %S
	 iManufacturer: %S
	 iSerialNumber: %S
```

compared to `diskutil info`

```
Disk Size:                 30.8 GB (30784094208 Bytes) (exactly 60125184 512-Byte-Units)
                                                                ^^^^^^^^
```

See `..183` sector count vs `..184` 512-Byte-Units

## TODO
- [x] changelog entry

---

## Checklist

---

> [!NOTE]
> **Low Risk**
> Single-path capacity normalization in the SCSI helper; improves correctness with minimal surface area, aside from rejecting oversized devices that previously got a bogus count.
> 
> **Overview**
> Fixes **READ CAPACITY(10)** handling in the USB host MSC SCSI layer: the response field is the **last accessible LBA**, not a sector count, so reported capacity was short by one block everywhere that value flowed (device info, FatFS `GET_SECTOR_COUNT`, blockdev size).
> 
> `scsi_cmd_read_capacity()` now names the field `last_lba`, converts to block count with **`last_lba + 1`**, and returns **`ESP_ERR_NOT_SUPPORTED`** when `last_lba == UINT32_MAX` (needs READ CAPACITY(16), not implemented). Changelog documents the fix under **[Unreleased]**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b20e7caccd52c9bc12921fc3e94c787bcf35c376. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tore-espressif (approved):
Thank you!

